PR #17, opened 2019-11-27
Schema.spaces now is optional

### LINKED ISSUE #14: schema.spaces table should be optional
`schema = {}` should be a valid notation

HustonMmmavr:
Close #14

### DIFF
--- a/ddl.lua
+++ b/ddl.lua
@@ -12,13 +12,14 @@ local function check_schema_format(schema)
         return nil, err_msg .. ' (schema expected, got ' .. type(schema) .. ')'
     end
 
-    if schema.spaces == nil then
-        return true
+    local spaces = schema.spaces
+    if spaces == nil then
+        spaces = {}
     end
 
-    if type(schema.spaces) ~= 'table' then
-        return nil, err_msg ..' invalid schema.spaces (optional table expected, got ' ..
-            type(schema.spaces) .. ')'
+    if type(spaces) ~= 'table' then
+        return nil, err_msg ..' invalid schema.spaces (?table expected, got ' ..
+            type(spaces) .. ')'
     end
 
     return true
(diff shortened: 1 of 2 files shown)


ddl.lua
```diff
@@ -12,8 +12,12 @@ local function check_schema_format(schema)
         return nil, err_msg .. ' (schema expected, got ' .. type(schema) .. ')'
     end
 
+    if schema.spaces == nil then
+        return true
+    end
+
     if type(schema.spaces) ~= 'table' then
-        return nil, err_msg ..' invalid schema.spaces (table expected, got ' ..
+        return nil, err_msg ..' invalid schema.spaces (optional table expected, got ' ..
```
rosik: `?table`

ddl.lua
```diff
@@ -76,6 +84,10 @@ local function set_schema(schema)
         return nil, string.format(err, 'set_schema')
     end
 
+    if schema.spaces == nil then
```
rosik: Same

ddl.lua
```diff
@@ -12,8 +12,12 @@ local function check_schema_format(schema)
         return nil, err_msg .. ' (schema expected, got ' .. type(schema) .. ')'
     end
 
+    if schema.spaces == nil then
+        return true
+    end
```
rosik:
```suggestion
local spaces = schema.spaces
if spaces == nil then
    spaces = {}
end
```

rosik: And replace all further mentions of `schema.spaces` to this local `spaces`

ddl.lua
```diff
@@ -26,6 +30,10 @@ local function check_schema(schema)
         return nil, string.format(err, 'check_schema')
     end
 
+    if schema.spaces == nil then
```
rosik: Same

PR #24, opened 2020-01-07
Add 'decimal' and 'double' to supported field types

knazarov:
Recent versions of Tarantool added support for 'decimal' and 'double'
field types to both Lua and SQL. Add them to DDL as well to support
creating spaces with those field types.

### DIFF
--- a/ddl/check.lua
+++ b/ddl/check.lua
@@ -52,6 +52,8 @@ local function check_field(i, field, space)
             array     = true,
             map       = true,
             any       = true,
+            decimal   = true,
+            double    = true,
         }
 
         if known_field_types[field.type] == nil then


knazarov:
@rosik what do you mean by `check that double field doesn't conflict with an index part of type "scalar"`?

rosik:
check out this test: https://github.com/tarantool/ddl/blob/ed058e0849333a76c720d1b872e46e38a7fbd8d0/test/check_schema_test.lua#L1127

olegrok:
> check out this test:

Please, don't test tarantool:
https://github.com/tarantool/tarantool/commit/d8193eb1c96955a20edf8421659d1ce53e63094e#diff-6e43339425d635084580e4acf5ea1201R131

It’s beyond the scope this module. The same you can copy-paste all test to check e.g. that unsigned is a scalar

rosik:
We don't test tarantool. DDL module doubles some checks (including this one):

https://github.com/tarantool/ddl/blob/ed058e0849333a76c720d1b872e46e38a7fbd8d0/ddl/check.lua#L278-L286

rosik:
And from your link I see there's much more internal logic behind field types, so maybe those doubled checks should be removed from the ddl module (we'll think about it later)

rosik:
Merged as #26

PR #35, opened 2020-07-15
Support fields of type uuid

### LINKED ISSUE #34: Support field type 'uuid' in schema
New field type 'uuid' was introduced in version 2.4, see the doc: https://www.tarantool.io/en/doc/2.4/reference/reference_lua/box_space/#lua-function.space_object.create_index

akudiyar:
Closes #34

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-## [1.2.0] - 2020-07-17
+## [1.2.0] - 2020-07-20
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [1.2.0] - 2020-07-17
+
+### Added
+
+- Support `uuid` types for tarantool 2.4
+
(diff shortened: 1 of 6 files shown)


olegrok:
See https://github.com/tarantool/ddl/pull/26

Changelog/Readme should be updated. And I assume that tests could be extended in the same way as it was done it MR above.

akudiyar:
> https://github.com/tarantool/ddl/blob/93d779e1950d67e964eb1bcf8a29d57be96d1e3b/ddl/check.lua#L283-L293
> 
> I think it should be added to the scalar types too.
> 
> (And tested appropriately)

UUID is not a scalar type, according to the sources:

image

olegrok:
> UUID is not a scalar type, according to the sources:

And tests https://github.com/tarantool/tarantool/commit/b238def8065d20070dcdc50b54c2536f1de4c7c7#diff-ac3bcfa197d7f09e185b4f93a4d8e869R1392

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [1.2.0] - 2020-07-17
```
olegrok, line 9: uuid type is unrealised yet

rosik: I'm OK to release it immediately.

rosik, line 9:
```suggestion
## [1.2.0] - 2020-07-20
```

PR #44, opened 2020-12-25
Use transactional ddl when applying schema

### LINKED ISSUE #46: Use transactional ddl when possible
Nowadays the DDL module performs `check_schema` by creating `_ddl_dummy` space and then dropping it. This approach has several drawbacks:

- It makes unnecessary transactions.
- It can't be applied on replicas (even writable), because it produces replication conflicts.

Let's wrap `check_schema` in `box.begin() / box.rollback()` and `ddl.set_schema` in `box.begin() / box.commit()`.

N.B. Tarantool 1.10 doesn't have transactional ddl, it'll only work in 2.x.

olegrok:
Physical creation of schema for test is the most painful part of
ddl module. DDL is replicated and it's quite awful when another
instances could see test schema modification.

This patch enables using of signle-yield transacational DDL
statements that allows to create empty spaces and indexes
transactionally. But since this schema modifications is needed
only for tests we rollback them at the end. So, another instances
don't see any test changes after this patch. For details see [1].

 [1] tarantool/tarantool#4083

Closes #46

### DIFF
--- a/ddl.lua
+++ b/ddl.lua
@@ -1,6 +1,7 @@
 local ddl_get = require('ddl.get')
 local ddl_set = require('ddl.set')
 local ddl_check = require('ddl.check')
+local ddl_db = require('ddl.db')
 local utils = require('ddl.utils')
 
 local function check_schema_format(schema)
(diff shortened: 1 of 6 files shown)


ddl.lua
```diff
@@ -87,24 +75,38 @@ local function check_schema(schema)
...
 
-    local ok, err = check_schema(schema)
-    if not ok then
-        return nil, err
+    if type(box.cfg) == 'function' then
+        return nil, "'box' module isn't configured yet"
+    end
+
+    if box.cfg.read_only then
+        return nil, "Instance is read-only (box.cfg.read_only=true)"
```
olegrok, line 89: It's not so. Instance could be in ro mode for several reasons: "orphan" mode or raft state is not a "leader".

olegrok: I suggest to change it to "see box.**info()**.read_only"

rosik (approved):

Issue #50, opened 2021-01-26
[2pt] Create a cartridge role for ddl.get_schema


akudiyar:
Related to https://github.com/tarantool/cartridge-java/issues/54

Consider a connector that needs to get the spaces schema from a Tarantool instance or a Cartridge cluster. In a Cartridge, we'll have to define a role containing the public API function for getting the schema. 

I suggest exposing the `get_schema` method in a public role like it is done with CRUD API (see https://github.com/tarantool/crud/pull/58).

rosik:
It's already there. See https://www.tarantool.io/en/doc/latest/book/cartridge/cartridge_api/modules/cartridge/#global-functions

```lua
_G.cartridge_get_schema()
_G.cartridge_set_schema(schema)
```

akudiyar:
Don't hurry up, we need to discuss this.

1) `cartridge.get_schema` returns schema in YAML format, which requires a connector to use a YAML library, which sometimes is not desired (I hope to not add the YAML library into cartridge-java at the moment since it is not needed for anything else yet). 
2) When using tarantool/migrations the `ddl.get_schema` API can be called from a router node, in this case that call will be more lightweight  than `cartridge.get_schema`
3) The output of `cartridge.get_schema` is not equal to `ddl.get_schema`. At least, the latter shows the internal `_bucket` space which may be useful in a connector.

PR #51, opened 2021-02-03
Create a cartridge role for ddl.get_schema. ddl-manager migration from cartridge to ddl repo

ArtDu:
# RFC: ddl-manager migration from cartridge to ddl repo. DDL  role for cartridge from box. 

* **Status**: In progress
* **Start date**: 03-02-2021
* **Issues**:  tarantool/ddl#50, tarantool/cartridge-java#54

## Summary
ddl-manager will locate with its parent(ddl). DDL get_schema role for cartridge from box.

## Background and motivation
Due to this migration, it is possible not to store two roles in parallel.
It is possible to simplify the code for the cartridge, using the logical ddl manager together with ddl itself.
DDL role does not need to be written manually by users, it comes out of the box along with the ddl package.
Possibility of accepting the scheme in yaml format and translating it into ddl format for users from box.

## Design Proposal
What to do with this code, just for the ddl manager it is not needed.
We are going to leave this code in the cartridge:
https://github.com/tarantool/ddl/blob/f010977d78b5bef362fe6a66ef798a36d9690445/cartridge/ddl-manager.lua#L101-L113

The idea is to make a call to ddl-manager functions from ddl api:
https://github.com/tarantool/ddl/blob/f010977d78b5bef362fe6a66ef798a36d9690445/ddl.lua#L149-L158

It is also worth renaming the function name to a logical format for translating yaml:
https://github.com/tarantool/ddl/blob/f010977d78b5bef362fe6a66ef798a36d9690445/cartridge/ddl-manager.lua#L57
https://github.com/tarantool/ddl/blob/f010977d78b5bef362fe6a66ef798a36d9690445/cartridge/ddl-manager.lua#L85

Instead of the entire config with the name of the section, etc., these functions will receive only the scheme.

### DIFF
--- a/test/integration/schema-api_role_test.lua
+++ b/test/integration/schema-api_role_test.lua
@@ -1,0 +1,172 @@
+local fio = require('fio')
+local t = require('luatest')
+local g = t.group('schema-api')
+
+local helpers = require('test.helper')
+
+math.randomseed(os.time())
+
+g.before_all(function()
+    g.cluster = helpers.Cluster:new({
+        datadir = fio.tempdir(),
+        server_command = helpers.entrypoint('schema-api_role_init'),
+        replicasets = {
+            {
+                uuid = helpers.uuid('a'),
+                alias = 'schema-api',
+                roles = { 'schema-api' },
+                servers = {
+                    { instance_uuid = helpers.uuid('a', 1), alias = 'schema-api' },
+                },
+            }
+        }
+    })
+    g.cluster:start()
+    g._section_name = 'schema.yml'
+    g._example_schema = [[
+     spaces:
+       customer:
+         engine: memtx
+         is_local: false
+         temporary: false
+         sharding_key: [customer_id]
+         format:
+           - {name: customer_id, type: unsigned, is_nullable: false}
+           - {name: bucket_id, type: unsigned, is_nullable: false}
+           - {name: fullname, type: string, is_nullable: false}
+         indexes:
+         - name: customer_id
+           unique: true
+           type: TREE
+           parts:
+             - {path: customer_id, type: unsigned, is_nullable: false}
+
+         - name: bucket_id
+           unique: false
+           type: TREE
+           parts:
+             - {path: bucket_id, type: unsigned, is_nullable: false}
+
+         - name: fullname
+           unique: true
+           type: TREE
+           parts:
+             - {path: fullname, type: string, is_nullable: false}
+    ]]
+    g.tbl_with_test_schema = {}
+    g.tbl_with_test_schema[g._section_name] = g._example_schema
+end)
+
+g.after_all(function()
+    g.cluster:stop()
+    fio.rmtree(g.cluster.datadir)
+end)
+
+g.test_ddl_get_schema = function()
+
+    local tbl = g.cluster.main_server.net_box:call("ddl.get_schema")
+    t.assert_equals(tbl['spaces'], {})
+
+    g.cluster.main_server.net_box:eval([[
+        box.schema.space.create("test1")
+    ]])
+    tbl = g.cluster.main_server.net_box:call("ddl.get_schema")
+    t.assert_equals(tbl['spaces'], {
+        test1 = {
+            engine = "memtx",
+            format = {},
+            indexes = {},
+            is_local = false,
+            temporary = false,
+        },
+    })
+
+    g.cluster.main_server.net_box:eval([[
+        box.schema.space.create("test2")
+    ]])
+    tbl = g.cluster.main_server.net_box:call("ddl.get_schema")
+    t.assert_equals(tbl['spaces'], {
+        test1 = {
+            engine = "memtx",
+            format = {},
+            indexes = {},
+            is_local = false,
+            temporary = false,
+        },
+        test2 = {
+            engine = "memtx",
+            format = {},
+            indexes = {},
+            is_local = false,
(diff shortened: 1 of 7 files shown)


test/integration/ddl-api_role_test.lua
```diff
@@ -0,0 +1,37 @@
...
+end)
+
+
+g.after_all(function()
+    g.cluster:stop()
+    fio.rmtree(g.cluster.datadir)
+end)
+
+g.test_ddl_schema = function()
+    local tbl = g.cluster.main_server.net_box:call("ddl.get_schema")
```
akudiyar, line 34: Let's add some spaces and validate if ddl do not always return empty table here

test/entrypoint/ddl-api_role_init.lua
```diff
@@ -0,0 +1,10 @@
+local cartridge = require('cartridge')
+local ok, err = cartridge.cfg({
+    workdir = 'tmp/db',
+    roles = {
+        'cartridge.roles.ddl-api'
+    },
+    cluster_cookie = 'ddl-cluster-cookie',
+})
+
+assert(ok, tostring(err))
```
akudiyar, line 10: Add newline (and to the other files too)

ddl.lua
```diff
@@ -146,8 +146,13 @@ local function get_schema()
...
 end
 
+local ddl_manager = require('cartridge.ddl-manager')
 return {
     check_schema = check_schema,
     set_schema = set_schema,
     get_schema = get_schema,
+    _section_name = ddl_manager._section_name,
+    _example_schema = ddl_manager._example_schema,
+    apply_schema_from_yaml = ddl_manager.apply_schema_from_yaml,
```
akudiyar, line 156: I thought again on the matter of the new "yaml" methods -- it seems not very comfortable for users. I think that supporting YAML in existing `set_schema` and `check_schema` methods is better (a YAML is a string, and now they support only a table as input).

cartridge/ddl-manager.lua
```diff
@@ -0,0 +1,128 @@
...
+end
+
+local function apply_schema_from_yaml(conf, opts)
+    checks('table', {is_master = 'boolean'})
+
+    if not opts.is_master then
+        return true
+    end
+
+    if conf[_section_name] == nil then
```
akudiyar, line 64: I think the section name is redundant here. The corresponding section as-is will be passed from Cartridge conf applier.

cartridge/ddl-manager.lua
```diff
@@ -0,0 +1,128 @@
...
+        )
+    end
+
+    return DecodeYamlError:pcall(yaml.decode, schema_yml)
+end
+
+local function apply_schema_from_yaml(conf, opts)
+    checks('table', {is_master = 'boolean'})
+
+    if not opts.is_master then
```
akudiyar, line 60: This check will be performed in Cartridge

cartridge/ddl-manager.lua
```diff
@@ -0,0 +1,128 @@
...
+        return nil, SetSchemaError:new(err)
+    end
+
+    return true
+end
+
+local function validate_schema_from_yaml(conf_new, _)
+    checks('table', 'table')
+
+    if conf_new[_section_name] == nil then
```
akudiyar, line 88: A YAML string will be passed, not conf section

cartridge/ddl-manager.lua
```diff
@@ -0,0 +1,128 @@
...
+    local schema, err = _from_yaml(conf_new[_section_name])
+    if schema == nil then
+        if err then
+            return nil, err
+        else
+            return true
+        end
+    end
+
+    if type(box.cfg) == 'function' then
```
akudiyar, line 101: Lines 101-113 must be removed, they will be left in Cartridge

cartridge/ddl-manager.lua
```diff
@@ -0,0 +1,128 @@
+local log = require('log')
+local ddl = require('ddl')
+local yaml = require('yaml')
+local checks = require('checks')
+local errors = require('errors')
+
+local failover = require('cartridge.failover')
```
akudiyar, line 7: The module should not be inside "cartridge" namespace, I suggest renaming it to `yaml-schema` or something relevant.

akudiyar (changes requested):

test/integration/schema-api_role_test.lua
```diff
@@ -0,0 +1,176 @@
...
+    local ok, err = g.cluster.main_server.net_box:eval([[
+        schema_api = require('cartridge.roles.schema-api')
+        return schema_api.validate_config(...)
+    ]], {g.tbl_with_test_schema, {}})
+    t.assert_equals(err, nil)
+end
+
+g.test_apply_config = function()
+
+
```
akudiyar, line 119: Extra newline

cartridge/ddl-manager.lua
akudiyar: Decided to leave the role code as it is in Cartridge

cartridge/ddl-manager.lua
akudiyar: Decided to leave as is

cartridge/ddl-manager.lua
akudiyar: Decided to leave as currently in Cartridge

akudiyar (approved):

PR #59, opened 2021-04-09
Add the ddl-manager from cartridge as a new role

### LINKED ISSUE #50: [2pt] Create a cartridge role for ddl.get_schema

rosik:
This is an extended version of #51.

* The role is renamed back to `ddl-manager`.
* Implement `stop()` callback for the sake of reloadability.
* Remove from CI the other `ddl` version installed by cartridge.
* Add the role to the cmake installer (it was missing in the original patch).
* Provide the API to be used in cartridge and test it a lot:

```lua
get_clusterwide_schema_yaml() -- return "spaces: {}"
get_clusterwide_schema_lua() -- return {spaces = {}}

set_clusterwide_schema_yaml("spaces: {}") -- return true
set_clusterwide_schema_lua({spaces = {}}) -- return true

check_schema_yaml("spaces: {}") -- return true
check_schema_lua({spaces = {}}) -- return true
```

Close #50

### DIFF
--- a/cartridge/roles/ddl-manager.lua
+++ b/cartridge/roles/ddl-manager.lua
@@ -1,0 +1,288 @@
+local log = require('log')
+local ddl = require('ddl')
+local yaml = require('yaml').new()
+local errors = require('errors')
+
+local cartridge = require('cartridge')
+local failover = require('cartridge.failover')
+
+local CheckSchemaError = errors.new_class('CheckSchemaError')
+
+yaml.cfg({
+    encode_use_tostring = true,
+    encode_load_metatables = false,
+    decode_save_metatables = false,
+})
+
+local _section_name = 'schema.yml'
+local _example_schema = [[## Example:
+#
+# spaces:
+#   customer:
+#     engine: memtx
+#     is_local: false
+#     temporary: false
+#     sharding_key: [customer_id]
+#     format:
+#       - {name: customer_id, type: unsigned, is_nullable: false}
+#       - {name: bucket_id, type: unsigned, is_nullable: false}
+#       - {name: fullname, type: string, is_nullable: false}
+#     indexes:
+#     - name: customer_id
+#       unique: true
+#       type: TREE
+#       parts:
+#         - {path: customer_id, type: unsigned, is_nullable: false}
+#
+#     - name: bucket_id
+#       unique: false
+#       type: TREE
+#       parts:
+#         - {path: bucket_id, type: unsigned, is_nullable: false}
+#
+#     - name: fullname
+#       unique: true
+#       type: TREE
+#       parts:
+#         - {path: fullname, type: string, is_nullable: false}
+]]
+
+local function apply_config(conf, opts)
+    if not opts.is_master then
+        return true
+    end
+
+    local schema_yml = conf[_section_name]
+    if schema_yml == nil then
+        return true
+    end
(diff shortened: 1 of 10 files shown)


cartridge/roles/ddl-manager.lua
```diff
@@ -0,0 +1,283 @@
...
+    if err ~= nil then
+        return nil, err
+    elseif schema_lua == nil then
+        return true
+    end
+
+    if not failover.is_leader() then
+        return cartridge.rpc_call(
+            'ddl-manager', 'check_schema_yaml', {schema_yml},
+            {prefer_local = true, leader_only = true}
```
olegrok, line 220: Current instance is not a leader why do you need "prefer_local = true" here?

cartridge/roles/ddl-manager.lua
```diff
@@ -0,0 +1,283 @@
...
+    apply_config = apply_config,
+
+    get_clusterwide_schema_yaml = get_clusterwide_schema_yaml,
+    get_clusterwide_schema_lua = get_clusterwide_schema_lua,
+
+    set_clusterwide_schema_yaml = set_clusterwide_schema_yaml,
+    set_clusterwide_schema_lua = set_clusterwide_schema_lua,
+
+    check_schema_yaml = check_schema_yaml,
+    check_schema_lua = check_schema_lua,
```
akudiyar, line 282: I think, it is not "lua" it is actually a json. So maybe it is worth adding "check_schema_json" and "check_schema" without suffix?

cartridge/roles/ddl-manager.lua
rosik: Fixed.

cartridge/roles/ddl-manager.lua
rosik: It's not a json, but a Lua table. And if you want to check json-encoded string you can use `_yaml` functions (because yaml is a superset of json format)

rosik (commented):
TODO:

- Export `CheckSchemaError`
- Distinguish it from other errors in `check_schema` call.

rosik:
> TODO:
> 
>     * Export `CheckSchemaError`
> 
>     * Distinguish it from other errors in `check_schema` call.

Done

test/helper.lua
```diff
@@ -0,0 +1,43 @@
+require('strict').on()
+
+local log = require('log')
+local digest = require('digest')
+local fio = require('fio')
+
+local ok, cartridge_helpers = pcall(require, 'cartridge.test-helpers')
+if not ok then
+    log.error('Please, install cartridge rock to run tests')
+    os.exit(1)
```
olegrok, line 10: Looks a bit strange. I suggest to skip in case if cartridge is not found.

rosik: Thanks for noticing. Fixed in fff3a7e1132b6e28b965ba383494b2d9664a4a59.

olegrok (approved):

PR #61, opened 2021-05-06
Store schema example in the clusterwide config

rosik:
If one applies an empty schema using the 'ddl-manager' role API, put in the clusterwide configuration an example instead.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Part of https://github.com/tarantool/cartridge/issues/1085
See also https://github.com/tarantool/cartridge/pull/1194

### DIFF
--- a/cartridge/roles/ddl-manager.lua
+++ b/cartridge/roles/ddl-manager.lua
@@ -4,7 +4,13 @@ local yaml = require('yaml').new()
 local errors = require('errors')
 
 local cartridge = require('cartridge')
+local twophase = require('cartridge.twophase')
 local failover = require('cartridge.failover')
+local vars = require('cartridge.vars').new('cartridge.roles.ddl-manager')
+
+vars:new('on_patch_trigger', nil)
+vars:new('_section_name', nil)
+vars:new('_example_schema', nil)
 
 local CheckSchemaError = errors.new_class('CheckSchemaError')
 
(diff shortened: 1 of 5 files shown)


test/role_test.lua
```diff
@@ -70,7 +70,7 @@ function g.test_yaml()
 
     t.assert_str_matches(call('get_clusterwide_schema_yaml'), '## Example:\n.+')
     t.assert_equals(call('get_clusterwide_schema_lua'), {spaces = {}})
-    t.assert_equals(get_section(), '')
+    t.assert_equals(call('get_clusterwide_schema_yaml'), get_section())
```
Steap2448, line 73: Does it make sense to check example explicitly?

rosik: Before this patch, the example was returned by `get_clusterwide_schema_yaml`, and the section was empty. Now I check that they're equal

cartridge/roles/ddl-manager.lua
```diff
@@ -47,12 +53,23 @@ local _example_schema = [[## Example:
 #         - {path: fullname, type: string, is_nullable: false}
 ]]
 
+-- Be gentle with cartridge.reload_roles
+twophase.on_patch(nil, vars.on_patch_trigger)
+function vars.on_patch_trigger(conf_new)
```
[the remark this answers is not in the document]
rosik: The trigger you've linked is the predecessor of the current one. I've copy-pasted it from there. While `cartridge-extensions` operates the whole directory `extensions/*`, `ddl` only handles a single section. And it's not going to handle any other in the foreseeable future.

That's why I decided to get rid of the `for` loop if it's what you've asked about.

Steap2448 (approved):

Issue #69, opened 2021-07-29
Use cache for database schema in get_schema()


ligurio:
CRUD will use DDL schema to obtain a sharding key using `get_schema()`. However, on every `get_schema()` call DDL module scan spaces to build the current database schema. Need to cache a database schema.

Totktonada:
Sadly, we have no public API for tracking a database schema version (without it there is no way to invalidate the cache, when some space or index is changed).

There is `box.internal.schema_version()`, which is obviously internal and has no guarantee that it'll exist in a future. It exists since 1.7.3-660-gcbfc235ee. We can take risks, use the internal function, file issue to expose it publicly (should be easy to implement and unlikely someone will be against). When it'll be exposed publicly, use the public function and fallback to the internal one if the public one does not exist.

PR #72, opened 2021-08-03
Add sharding_func to ddl

### LINKED ISSUE #71: Add sharding function section to ddl schema
RFC - https://github.com/tarantool/crud/issues/27#issuecomment-885849411

AnaNek:
Closes  #71

RFC - https://github.com/tarantool/crud/issues/27#issuecomment-885849411


ddl/get.lua
```diff
@@ -52,12 +52,31 @@ local function _get_index(box_space, box_index)
...
-    if box.space._ddl_sharding_key == nil then
+local function get_metadata(space_name, metadata_name)
+    local metadata_space_name = string.format("_ddl_%s", metadata_name)
+
+    if box.space[metadata_space_name] == nil then
+        return nil
+    end
+
+    return box.space[metadata_space_name]:get{space_name}
+end
```
ligurio, line 63: It would be good to think about cache metadata and its invalidation on schema update.

test/check_sharding_data_test.lua
```diff
@@ -88,6 +88,47 @@ function g.test_sharding_key_ok()
...
+    g.space.sharding_func = 'vshard.router.bucket_id_mpcrc32()'
+    local res, err = ddl.check_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(res, true)
+
+    -- sharding_func defined by user
+    g.space.sharding_func = {body = 'function(key) return <...> end'}
+    local res, err = ddl.check_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(res, true)
```
ligurio, line 128: Testing `check_schema()` is good but it is not enough. I propose to test `set_schema()` too.

ligurio (changes requested):
Thanks for patches! see my comments inline.

README.md
```diff
@@ -37,6 +37,26 @@ Additionally when a section `sharding_key` is not empty in YAML schema call of
 function `ddl.set_schema(schema)` creates a space `_ddl_sharding_key` with two
 fields: `space_name` with type `string` and `sharding_key` with type `array`.
 
+Similarly for `sharding_func`: when a section `sharding_func` is not empty in YAML schema call of
+function `ddl.set_schema(schema)` creates a space `_ddl_sharding_func` with three
+fields: `space_name` with type `string`, `sharding_func` with type `string` and
+`lua_code` with type `boolean`.
```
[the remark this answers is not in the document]
AnaNek: I add `lua_code` field to accept and return `sharing_func` in the same format like if user would like to use custom sharding function he would write:

```lua
sharding_func = {
  body = 'function(key) return <...> end'
}
```

And `get_schema()` would return not just `sharding_func = 'function(key) return <...> end'` but table with 'body' section. But maybe it is okay if `get_schema()` returns `sharding_func = 'function(key) return <...> end'`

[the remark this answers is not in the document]
ligurio: > A schema provider may expose additional functions, which allow to set particular fields:
> 
> ```
> -- Set sharding function for a space.
> --
> -- func: 'mpcrc32' | 'strcrc32' | 'dot.notation.func' | {body = 'function(key) return <...> end'}
> <schema_provider>.set_sharding_func(space_name, func)
> ```
> However it is generally out of scope of this proposal.

test/check_sharding_data_test.lua
```diff
@@ -88,6 +88,47 @@ function g.test_sharding_key_ok()
...
+    g.space.sharding_func = 'strcrc32'
+    local res, err = ddl.check_schema(schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(res, true)
+
+    -- vshard.router.bucket_id_strcrc32()
+    g.space.sharding_func = 'vshard.router.bucket_id_strcrc32()'
+    local res, err = ddl.check_schema(schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(res, true)
```
ligurio, line 110: One can pass any custom function using 'dot.notation' so I would create a test for a custom sharding function and specify it using 'dot.notation'.

test/set_sharding_data_test.lua
```diff
@@ -53,6 +53,12 @@ local sharding_key_format = {
     {name = 'sharding_key', type = 'array', is_nullable = false}
 }
 
+local sharding_func_format = {
+    {name = 'space_name', type = 'string', is_nullable = false},
+    {name = 'sharding_func', type = 'string', is_nullable = false},
+    {name = 'lua_code', type = 'boolean', is_nullable = false}
+}
+
```
ligurio, line 61: I would describe a schema for a space with sharding functions in a README and say that space format is a part of DDL API.

ligurio (changes requested):

ddl/get.lua
AnaNek: Updated

test/check_sharding_data_test.lua
AnaNek: Please have a look at `set_sharding_func_test.lua`, there are tests with `set_schema()`

test/check_sharding_data_test.lua
AnaNek: Updated: in test `test_sharding_func_ok` in file `check_sharding_data_test.lua`:

```lua
-- sharding_func defined by user and set in _G
    local user_sharding_func = 'user_sharding_func'
    rawset(_G, user_sharding_func, {})
    g.space.sharding_func = user_sharding_func
    local res, err = ddl.check_schema(g.schema)
    t.assert_equals(err, nil)
    t.assert_equals(res, true)
```

In test `test_invalid_sharding_func` in file `check_sharding_data_test.lua` there is check that tests case with identifier (`mpcrc32`/`strcrc32`) and case with user function in `_G`:

```lua
-- unknown name
    g.space.sharding_func = 'bad_sharding_func'
    local ok, err = ddl.check_schema(g.schema)
    t.assert_equals(err,
        [[spaces["space"].sharding_func: unknown sharding_func "bad_sharding_func"]]
    )
    t.assert_equals(ok, nil)
```

In file `set_sharding_data_test.lua` test `test_ddl_sharding_func_space_user_func_name` was added

test/set_sharding_data_test.lua
AnaNek: Updated

ddl/check.lua
```diff
@@ -628,75 +628,113 @@ local function check_index(i, index, space)
...
+
+            if type(space.sharding_func.body) ~= 'string' then
+                return nil, string.format(
+                    "spaces[%q].sharding_func.body: bad value" ..
+                    " (string expected, got %s)",
+                    space.name, type(space.sharding_func.body)
+                )
+            end
+        end
+    end
```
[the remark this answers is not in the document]
AnaNek: Not done because of the security

README.md
```diff
@@ -37,6 +37,26 @@ Additionally when a section `sharding_key` is not empty in YAML schema call of
...
+
+These sharding functions are supported now:
+
+- `strcrc32`;
+- `mpcrc32`.
```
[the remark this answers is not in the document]
ligurio: Earlier in a conversation in a team's chat we decided to combine a way to specify vshard's sharding functions using constants with a way using `dot.notation`. It allows to reduce a code a bit and have unified approach.
I see that README and code (at least in `ddl/check.lua`) contains `strcrc32` and `mpcrc32`...

ddl/check.lua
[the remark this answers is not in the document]
ligurio: Well, imagine user prepared a DDL schema and accidentally added a mistake to a function name.
`ddl.set_schema()` will report that schema successfully installed, but then sharding will be malfunction due to mistake...

README.md
```diff
@@ -37,6 +37,26 @@ Additionally when a section `sharding_key` is not empty in YAML schema call of
...
+
+You can use short names like `strcrc32`/`mpcrc32` 
+or full names `vshard.router.bucket_id_strcrc32()`/`vshard.router.bucket_id_mpcrc32()`.
+
+Also you can pass your own sharding function using lua code:
+```lua
+sharding_func = {
+  body = 'function(key) return <...> end'
+}
```
ligurio, line 60: We need to explicitly specify what sharding function can accept (number of parameters and their types) and what function must return (number of parameters and their types).

ligurio (changes requested):

README.md
AnaNek: Updated

test/check_sharding_data_test.lua
```diff
@@ -219,3 +268,30 @@ function g.test_invalid_key_reference_path()
...
+    t.assert_equals(ok, nil)
+
+    -- sharding_func without body
+    g.space.sharding_func = {}
+    local ok, err = ddl.check_schema(g.schema)
+    t.assert_equals(err,
+        [[spaces["space"].sharding_func: sharding_func exists in the space as table,]] ..
+        [[but there's no 'body' field in sharding_func section]]
+    )
+    t.assert_equals(ok, nil)
```
ligurio, line 296: I would split test into a number of tests.  `assert()` failed in the beginning of test will block other testcases.

ligurio (changes requested):

test/check_sharding_data_test.lua
```diff
@@ -88,6 +88,55 @@ function g.test_sharding_key_ok()
...
+    t.assert_equals(res, true)
+
+    -- mpcrc32
+    g.space.sharding_func = 'mpcrc32'
+    local res, err = ddl.check_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(res, true)
+
+    -- vshard.router.bucket_id_mpcrc32()
+    g.space.sharding_func = 'vshard.router.bucket_id_mpcrc32()'
```
Totktonada, line 119: Please, use just name, without parentheses.

README.md
```diff
@@ -37,6 +37,39 @@ Additionally when a section `sharding_key` is not empty in YAML schema call of
...
+```
+
+Defined function must have a prototype according to the following rules:
+
+**Parameters**
+
+key (number | table | string) – a hash key.
```
Totktonada, line 67: Why hash? I would say, 'a sharding key'.

README.md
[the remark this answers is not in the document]
Totktonada: I would follow `box.space._func` format and name it just 'body', not 'lua_body'.

README.md
```diff
@@ -141,6 +174,8 @@ format = {
                 -- unsharded spaces must NOT have
                 -- field and index named 'bucket_id'
             },
+            sharding_func = 'mpcrc32' | 'strcrc32' | 'dot.notation.func' |
+                            {body = 'function(key) return <...> end'} | <...>,
```
Totktonada, line 178: The last `<...>` looks redundant and maybe even confusing.

ddl/check.lua
```diff
@@ -628,75 +628,113 @@ local function check_index(i, index, space)
...
+
+    return true
+end
+
+local function check_sharding_key(space)
```
Totktonada, line 733: NB: Now the function does nothing when there is no `space.sharding_key`. Before the change it did check that there is no `bucket_id` space and index. I understand the change: it is to don't do the same check twice.

Tehcnically this function is available for a user, but the ddl submodules are not mentioned in the README, so it is the grey zone. I would consider them all as internal (so the change is okay, at least from the compatibility side of view).

Anyway, commented it here to don't forget to look again.

ddl/check.lua
```diff
@@ -628,75 +628,113 @@ local function check_index(i, index, space)
...
-            return nil, string.format(
-                "spaces[%q].indexes[%q]: bucket_id is used for sharding, " ..
-                "but there's no spaces[%q].sharding_key defined",
-                space.name, 'bucket_id', space.name
-            )
-        end
-        return true
-    end
-
+local function check_bucket_id_is_valid(space_name, bucket_id_field, bucket_id_index)
```
Totktonada, line 631: Can we pass just `space` object instead of those three variables? It would make the code a bit easier: `space.fields.bucket_id` and `space.indexes.bucket_id` are more or less obvious from the name, while, say, `bucket_id_field` may be interpreted as a tuple field value or as a tuple field descriptor.

BTW, the idea of the method is to check two different things: correctness of the 'bucket_id' field in a sharded space and correctness of the 'buckte_id' index in a sharded space. They're look strange under one function and the name of the function does not reveal the idea (just because it is hard to do so in a few words). I hope there is away to arrange this code better.

ddl/check.lua
```diff
@@ -746,6 +784,47 @@ local function check_sharding_key(space)
     return true
 end
 
+local function check_sharding_data(space)
```
Totktonada, line 787: I propose to call sharding key descriptor, sharding function, descriptors of the bucket_id field and the bucket_id index 'sharding metatada' (not 'data'). At least to be consistent in the terminology. But it also looks more natural: the whole database schema is data about data stored in the database: the database metadata.

ddl/check.lua
```diff
@@ -628,75 +628,113 @@ local function check_index(i, index, space)
...
+                ['strcrc32'] = true,
+                ['vshard.router.bucket_id_strcrc32()'] = true,
+                ['vshard.router.bucket_id_mpcrc32()'] = true,
+            }
+            if not known_sharding_funcs[space.sharding_func] and rawget(_G, space.sharding_func) == nil then
+                return nil, string.format(
+                    "spaces[%q].sharding_func: unknown sharding_func %q",
+                    space.name, space.sharding_func
+                )
+            end
```
Totktonada, line 710: How about a user provided function that is set by its name?

ddl/set.lua
```diff
@@ -77,10 +79,22 @@ local function create_space(space_name, space_schema, opts)
...
 
     if not is_dummy then
-        local ok, err = pcall(create_sharding_key, space_name, space_schema)
+        local ok, err = pcall(create_metadata, space_name, {space_schema.sharding_key}, "sharding_key")
         if not ok then
             return nil, string.format("spaces[%q].sharding_key: %s", space_name, err)
         end
+
+        local sharding_func = space_schema.sharding_func
+        local lua_code = false
```
Totktonada, line 88: I would write a sharding function name into one field and sharding function body into another one. It seems logical to make both fields nullable. Mixing names and bodies into one field looks error prone for me.

ddl/get.lua
```diff
@@ -52,12 +52,51 @@ local function _get_index(box_space, box_index)
...
+    end
+
+    local record = box.space[metadata_space_name]:get{space_name}
+    if index then
+        metadata_cache[index] = metadata_cache[index] or {}
+        metadata_cache[index][metadata_name] = record
+    end
+
+    return record
+end
```
Totktonada, line 83: I vote for don't do caching here. There are several thoughts around:

1. Now the module constructs the schema each time it is requested. So frequent access assumes caller side caching anyway. If we'll implement caching within the module (as proposed in #69), we'll likely do it on a higher level.
2. There are several thoughts why this caching algorithm may not work as expected:
    - It depends on when the garbage collector will delete an index object. So even if the space is deleted and created again we can return data from the cache.
    - Even if we would call `collectgarbage()` manually (that would be strange but anyway), it does not protect us from the situation, when a user kept a link to the old index object.
    - The modification of the space (inserting, removing, updating tuples) will not loss a reference to the index object. So, again, we can return obsolete data.

test/check_sharding_data_test.lua
Totktonada: Looks resolved. @ligurio?

test/set_sharding_data_test.lua
Totktonada: Seems resolved. @ligurio?

README.md
Totktonada: It looks resolved. @ligurio?

README.md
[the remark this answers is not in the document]
Totktonada: To Sergey's comment about `set_sharding_func()`. Sure, I would leave it out of scope here.

I'm a bit worry about lack of built-in ability to add a sharding function to an existing space. But I think that the main (not necessary the only) way to do so should be passing an updated schema into `set_schema()`. The module is not ready for this kind of operations: AFAIS the `set_schema()` call always skips existing spaces.

We have the space format described in the README, so the workaround exists: one can add a sharding function manually.

Totktonada:
It seems, that's all from me on this review iteration.

README.md
```diff
@@ -37,6 +37,26 @@ Additionally when a section `sharding_key` is not empty in YAML schema call of
 function `ddl.set_schema(schema)` creates a space `_ddl_sharding_key` with two
 fields: `space_name` with type `string` and `sharding_key` with type `array`.
 
+Similarly for `sharding_func`: when a section `sharding_func` is not empty in YAML schema call of
```
ligurio, line 40: AFAIR current version of code creates a space even user didn't specify `sharding_func` section in a schema. And I think it's a preferred behavior.

ligurio (changes requested):

README.md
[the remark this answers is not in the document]
AnaNek: Updated:

`_ddl_sharding_func` space format now contains `sharding_func_name` and `sharding_func_body` nullable fields.
```lua
local function _set_schema(schema)
    ...
    set_metadata_space(
        'sharding_func',
        {
            {name = 'space_name', type = 'string', is_nullable = false},
            {name = 'sharding_func_name', type = 'string', is_nullable = true},
            {name = 'sharding_func_body', type = 'string', is_nullable = true},
        }
    )
    ...
end
```

README.md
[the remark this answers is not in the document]
AnaNek: Updated:

There is `check_custom_sharding_func_is_valid` function now to check correctness of user sharding function.

```lua
local function check_custom_sharding_func_is_valid(sharding_func_name)
    local sharding_func = rawget(_G, sharding_func_name)
    if sharding_func == nil then
        return false
    end

    if type(sharding_func) == "function" then
        return true
    end

    if getmetatable(sharding_func) and getmetatable(sharding_func).__call then
        return true
    end

    return false
end
```

ddl/check.lua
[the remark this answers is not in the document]
AnaNek: Updated:

There is `check_custom_sharding_func_is_valid` function now to check correctness of user sharding function.

```lua
local function check_custom_sharding_func_is_valid(sharding_func_name)
    local sharding_func = rawget(_G, sharding_func_name)
    if sharding_func == nil then
        return false
    end

    if type(sharding_func) == "function" then
        return true
    end

    if getmetatable(sharding_func) and getmetatable(sharding_func).__call then
        return true
    end

    return false
end
```

ddl/get.lua
AnaNek: Cache for metadata was removed for some reasons that was described by @Totktonada (https://github.com/tarantool/ddl/pull/72#discussion_r704860471)

test/check_sharding_data_test.lua
AnaNek: Done.

test/check_sharding_data_test.lua
AnaNek: Updated:

In `ddl/check.lua`
```lua
local function check_sharding_func(space)
    if not space.sharding_func then
        return true
    end

    do -- check that sharding_func format is valid
        ...
        if type(space.sharding_func) == 'string' then
            local known_sharding_funcs = {
                ['vshard.router.bucket_id_strcrc32'] = true,
                ['vshard.router.bucket_id_mpcrc32'] = true,
            }
        ...
        end
    ...
    end
...
end
```

And in `test/`

```lua
function g.test_vshard_sharding_func()
    -- vshard.router.bucket_id_strcrc32
    g.space.sharding_func = 'vshard.router.bucket_id_strcrc32'
    local res, err = ddl.check_schema(g.schema)
    t.assert_equals(err, nil)
    t.assert_equals(res, true)

    -- vshard.router.bucket_id_mpcrc32
    g.space.sharding_func = 'vshard.router.bucket_id_mpcrc32'
    local res, err = ddl.check_schema(g.schema)
    t.assert_equals(err, nil)
    t.assert_equals(res, true)
end
```

README.md
AnaNek: Updated

README.md
AnaNek: Updated

ddl/check.lua
AnaNek: I change `check_sharding_key` and `check_sharding_func` functions visibility. Now both of them are internal:

```
     check_index = check_index,
-    check_field = check_field,
-    check_sharding_key = check_sharding_key,
-    check_sharding_func = check_sharding_func
+    check_field = check_field
 }
```

Should I change other check functions (`check_index` etc.) to internal too?

ddl/check.lua
AnaNek: I split `check_bucket_id_is_valid` to `check_bucket_id_field_value(space)` and `check_bucket_id_index(space)`

ddl/check.lua
AnaNek: Updated:

```lua
local function check_sharding_metadata(space)
...
end
```

ddl/check.lua
AnaNek:
```lua
if type(space.sharding_func) == 'string' then
    local known_sharding_funcs = {
        ['vshard.router.bucket_id_strcrc32'] = true,
        ['vshard.router.bucket_id_mpcrc32'] = true,
    }
    if not known_sharding_funcs[space.sharding_func]
    and not check_custom_sharding_func_is_valid(space.sharding_func) then
        return nil, string.format(
            "spaces[%q].sharding_func: unknown sharding_func %q",
            space.name, space.sharding_func
        )
    end
end
```

There is `check_custom_sharding_func_is_valid` function now to check correctness of user sharding function.
If function name provided by user is set in `_G` and this object has type `function` or method `__call` then function name is correct.

```lua
local function check_custom_sharding_func_is_valid(sharding_func_name)
    local sharding_func = rawget(_G, sharding_func_name)
    if sharding_func == nil then
        return false
    end

    if type(sharding_func) == "function" then
        return true
    end

    if getmetatable(sharding_func) and getmetatable(sharding_func).__call then
        return true
    end

    return false
end
```

ddl/set.lua
AnaNek: Updated:

`_ddl_sharding_func` space format now contains `sharding_func_name` and `sharding_func_body` nullable fields.
```lua
local function _set_schema(schema)
    ...
    set_metadata_space(
        'sharding_func',
        {
            {name = 'space_name', type = 'string', is_nullable = false},
            {name = 'sharding_func_name', type = 'string', is_nullable = true},
            {name = 'sharding_func_body', type = 'string', is_nullable = true},
        }
    )
    ...
end
```

ddl/get.lua
AnaNek: Metadata cache has been removed

README.md
AnaNek: Updated

Note: I guess same changes are fair for `_ddl_sharding_key` space info too

```
-Additionally when a section `sharding_key` is not empty in YAML schema call of
-function `ddl.set_schema(schema)` creates a space `_ddl_sharding_key` with two
+Call of function `ddl.set_schema(schema)` creates a space `_ddl_sharding_key` with two
 fields: `space_name` with type `string` and `sharding_key` with type `array`.
 
-Similarly for `sharding_func`: when a section `sharding_func` is not empty in YAML schema call of
-function `ddl.set_schema(schema)` creates a space `_ddl_sharding_func` with three
-fields: `space_name` with type `string`, `sharding_func_name` with type `string` and
-`sharding_func_body` with type `string`.
+Similarly for `sharding_func`: call of function `ddl.set_schema(schema)` creates
+a space `_ddl_sharding_func` with three fields: `space_name` with type `string`,
+`sharding_func_name` with type `string` and `sharding_func_body` with type `string`.
```

ligurio: > Note: I guess same changes are fair for _ddl_sharding_key space info too

it is better to make a separate patch with this change

ddl/check.lua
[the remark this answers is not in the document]
ligurio: `check_custom_sharding_func_is_valid` checks a number of properties for sharding function, but not input and output data.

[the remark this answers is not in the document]
ligurio: I'll add such check in in scope of https://github.com/tarantool/ddl/issues/76

ddl/check.lua
```diff
@@ -628,75 +628,131 @@ local function check_index(i, index, space)
...
+            "spaces[%q].indexes[%q].parts[1].path: invalid field reference " ..
+            "(reference to bucket_id expected, got %s)",
+            space.name, 'bucket_id', space.indexes['bucket_id'].parts[1].path
+        )
+    end
+
+    return true
+end
+
+local function check_custom_sharding_func_is_valid(sharding_func_name)
```
ligurio, line 684: All functions in this module check data for validness. But don't use postfix `_is_valid` in name. I think we can follow this rules here too.

ligurio (changes requested):

README.md
AnaNek: Add in #78

ddl/check.lua
AnaNek: Updated.

test/set_sharding_data_test.lua
```diff
@@ -244,3 +293,136 @@ function g.test_ddl_sharding_key_space()
...
+        {
+            {'space_one', space_one.sharding_func}
+        }
+    )
+
+    local ddl_schema = ddl.get_schema()
+    t.assert_equals(ddl_schema, schema)
+end
+
+function g.test_ddl_user_sharding_func_call_method()
```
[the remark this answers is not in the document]
AnaNek: Added.

test/check_sharding_data_test.lua
```diff
@@ -87,6 +87,47 @@ function g.test_sharding_key_ok()
...
+function g.test_user_sharding_func_by_name()
+    local user_sharding_func_name = 'user_sharding_func'
+    rawset(_G, user_sharding_func_name, function(key) return key end)
+    g.space.sharding_func = user_sharding_func_name
+    local res, err = ddl.check_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(res, true)
+end
+
+function g.test_user_sharding_func_call_method()
```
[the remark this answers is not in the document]
AnaNek: Added.

README.md
```diff
@@ -33,10 +33,39 @@ in [documementation](https://www.tarantool.io/en/doc/latest/book/box/data_model/
...
+Also, you can pass your own sharding function by defining the function name in `_G`
+or by specifying lua code directly in `body` field:
+```lua
+sharding_func = {
+  body = 'function(key) return <...> end'
+}
+```
+
+Your defined sharding function in `_G` should have type `function`
+or type `table` with `__call` method.
```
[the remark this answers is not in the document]
AnaNek: Updated:

```lua
local function is_callable(object)
    if type(object) == 'function' then
        return true
    end

    if type(object) == 'cdata' then
        return true
    end

    if type(object) == 'userdata' then
        return true
    end

    local object_metatable = getmetatable(object)
    if type(object) == 'table' and object_metatable
    and object_metatable.__call then
        return type(object_metatable.__call) == 'function'
    end

    return false
end
```

README.md
```diff
@@ -33,10 +33,39 @@ in [documementation](https://www.tarantool.io/en/doc/latest/book/box/data_model/
...
     Return values: `true` if no error, otherwise return `nil, err`
 
-Additionally when a section `sharding_key` is not empty in YAML schema call of
-function `ddl.set_schema(schema)` creates a space `_ddl_sharding_key` with two
+Call of function `ddl.set_schema(schema)` creates a space `_ddl_sharding_key` with two
 fields: `space_name` with type `string` and `sharding_key` with type `array`.
 
+Similarly for `sharding_func`: call of function `ddl.set_schema(schema)` creates
+a space `_ddl_sharding_func` with three fields: `space_name` with type `string`,
+`sharding_func_name` with type `string` and `sharding_func_body` with type `string`.
```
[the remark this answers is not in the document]
AnaNek: Done:

```
Similarly for `sharding_func`: call of function `ddl.set_schema(schema)` creates
a space `_ddl_sharding_func` with three fields: `space_name`, `sharding_func_name` and
`sharding_func_body` with type string.
```

ligurio (changes requested):

ddl/check.lua
```diff
@@ -628,75 +628,141 @@ local function check_index(i, index, space)
...
             )
         end
 
-        if space.indexes.bucket_id.unique then
-            return nil, string.format(
-                "spaces[%q].indexes[%q].unique: bucket_id index can't be unique",
-                space.name, 'bucket_id'
-            )
+        if type(space.sharding_func) == 'string' then
+            if not space.sharding_func:match('^[%w_]+[%.[%w_]+%.]*[%w_]+$')
```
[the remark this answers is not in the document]
AnaNek: Comment in code added. This check is for dot.notation: sharding func name should not begin with dot or digit and last symbol may not be dot. If sharding func name  has at least one dot according regex then it is a dot.notation and we don't check this name in `_G` and checking of possibility of a call  happens in #80 , otherwise this is a name of user sharding func that should be checked in `_G`.

ddl/check.lua
```diff
@@ -628,75 +628,141 @@ local function check_index(i, index, space)
...
+            "(reference to bucket_id expected, got %s)",
+            space.name, 'bucket_id', space.indexes['bucket_id'].parts[1].path
+        )
+    end
+
+    return true
+end
+
+local function is_callable(object)
+    if type(object) == 'function' then
```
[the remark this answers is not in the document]
ligurio: Thanks!

README.md
```diff
@@ -39,6 +39,38 @@ in [documementation](https://www.tarantool.io/en/doc/latest/book/box/data_model/
...
+Also, you can pass your own sharding function by defining the function name in `_G`
+or by specifying lua code directly in `body` field:
+```lua
+sharding_func = {
+  body = 'function(key) return <...> end'
+}
+```
+
+Your defined sharding function in `_G` should have type `function`
+or `table` | `cdata` | `userdata` with `__call` method.
```
Totktonada, line 62: Nit: method -> metamethod.

Totktonada:
(I'm still in progress, please, don't take it as a full review iteration.)

ddl/check.lua
```diff
@@ -628,75 +628,151 @@ local function check_index(i, index, space)
...
-                "spaces[%q].format[%q].type: bad value (unsigned expected, got %s)",
-                space.name, 'bucket_id', space.fields.bucket_id.type
-            )
-        end
+    -- all objects with type `userdata` are allowed
+    -- because there is no easy way to get
+    -- metatable.__call of object with type `userdata`
+    if type(object) == 'userdata' then
+        return true
+    end
```
Totktonada, line 701: `getmetatable()` works for a userdata in the same way as for a table. However there is a weird case about both userdata and table objects: if `__metatable` field is defined within a metatable, we unable to get actual metatable to check its `__call` field.

```lua
tarantool> x = setmetatable({}, {__call = function(self, ...) return ... end, __metatable = 'hidden'})
tarantool> getmetatable(x)
---
- hidden
...
tarantool> x(1, 2, 3)
---
- 1
- 2
- 3
...
```

However there is a trick: we can use `debug.getmetatable()`.

```lua
tarantool> debug.getmetatable(x)
---
- __metatable: hidden
  __call: 'function: 0x403d1130'
...
```

It is not nice to rely on the 'debug' library in production code. I'll cite Mike Pall (from [here](https://github.com/LuaJIT/LuaJIT/issues/536#issuecomment-565797780)):

> The purpose of the `debug.*` functions is for debugging Lua code. Period. In general, you'd want to turn the JIT compiler off for that, because it does a lot of things that will interfere with that task.

I would consider table and userdata objects with a 'protected' metatable as non-callables and leave a comment about this case.

ddl/set.lua
```diff
@@ -31,12 +31,14 @@ local function create_index(box_space, ddl_index)
...
-    if not space.sharding_key then
+local function create_metadata(space_name, metadata, metadata_name)
+    if not next(metadata) then
         return
     end
 
-    box.space._ddl_sharding_key:insert{space_name, space.sharding_key}
+    local metadata_space_name = string.format("_ddl_%s", metadata_name)
+
+    box.space[metadata_space_name]:insert{space_name, unpack(metadata)}
```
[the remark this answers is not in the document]
AnaNek: Updated:

```diff
local function create_metadata(space_name, metadata, metadata_name)
 
     local metadata_space_name = string.format("_ddl_%s", metadata_name)
 
-    box.space[metadata_space_name]:insert{space_name, unpack(metadata)}
+    box.space[metadata_space_name]:insert{space_name, unpack(metadata, 1, table.maxn(metadata))}
 end
```

README.md
AnaNek: Fixed.

README.md
```diff
@@ -39,6 +39,38 @@ in [documementation](https://www.tarantool.io/en/doc/latest/book/box/data_model/
 Call of function `ddl.set_schema(schema)` creates a space `_ddl_sharding_key` with two
 fields: `space_name` with type `string` and `sharding_key` with type `array`.
 
+Similarly for `sharding_func`: call of function `ddl.set_schema(schema)` creates
+a space `_ddl_sharding_func` with three fields: `space_name`, `sharding_func_name` and
+`sharding_func_body` with type string.
+
+If you want to use sharding function from some module, you need to require the module
+with sharding function first. For example: to use sharding functions like
```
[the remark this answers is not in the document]
AnaNek: Updated:

```
If you want to use sharding function from some module, you need to require
and set to `_G` the module with sharding function first.
```

README.md
```diff
@@ -39,6 +39,38 @@ in [documementation](https://www.tarantool.io/en/doc/latest/book/box/data_model/
...
+```
+
+Your defined sharding function in `_G` should have type `function`
+or `table` | `cdata` | `userdata` with `__call` method.
+
+Defined function must have a prototype according to the following rules:
+
+**Parameters**
+
+key (number | table | string) – a sharding key.
```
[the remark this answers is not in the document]
AnaNek: I guess yes, because `vshard` sharding functions accept any Lua object as key as described in [docs](https://www.tarantool.io/en/doc/latest/reference/reference_rock/vshard/vshard_api/#router-api-bucket-id-strcrc32)

ddl/check.lua
AnaNek: Right now check for `userdata` is updated:

```lua
    local object_metatable = getmetatable(object)
    if (type(object) == 'table' or type(object) == 'userdata')
    and object_metatable and object_metatable.__call then
        -- `__call` metamethod can be only the `function`
        -- and cannot be a `table` | `userdata` | `cdata`
        -- with `__call` methamethod on its own
        return type(object_metatable.__call) == 'function'
    end
```

ddl/check.lua
```diff
@@ -628,75 +628,147 @@ local function check_index(i, index, space)
...
+            -- sequence of letters, digits, or underscore symbols,
+            -- and may not begin with a digit or dot, the last symbol
+            -- may not be a dot
+            if not space.sharding_func:match('^[%a_]%w*[%.[%w_]+%.]*[%w_]+$')
+            and not check_custom_sharding_func(space.sharding_func) then
+                return nil, string.format(
+                    "spaces[%q].sharding_func: unknown sharding_func %q",
+                    space.name, space.sharding_func
+                )
+            end
```
Totktonada, line 742: (I wrote the text below during thinking, so a lot of thoughts around are there. You can skip it, the summary is at the bottom.)

First of all, it seems, the regex is broken by square brackets within square brackets:

```lua
tarantool> ('xxx'):match('^[%a_]%w*[%.[%w_]+%.]*[%w_]+$')
---
- null
...
```

Or we check one component names separately and call them 'custom'? Consider more examples:

```lua
tarantool> ('foo..bar'):match('^[%a_]%w*[%.[%w_]+%.]*[%w_]+$')
---
- foo..bar
...
tarantool> ('foo.ba[r.baz'):match('^[%a_]%w*[%.[%w_]+%.]*[%w_]+$')
---
- foo.ba[r.baz
...
```

(It seems that Lua regexps are not powerful enough for validation of dot notation, see below.)

Next, LuaJIT accepts some ASCII symbols and any correct UTF-8 multibyte sequences (and some incorrect ones) as indentifier symbols. The token should not start from a digit and should not be a reserved word (like 'end'). See `lj_char_isident()`, `lj_lex()` and `TKDEF` macro at top of the `lj_lex.h` file.

Tarantool requires correct printable UTF-8 characters in `box.schema.func.create()`, but not necessarily an identifier.

See examples:

```lua
$ echo -en 'local function \xc0\x61\x80\x61(...) print(...) end \xc0\x61\x80\x61(1,2,3)' | tarantool
1	2	3
```

```lua
tarantool> ok, err = pcall(box.schema.func.create, '\xc0\x61\x80\x61')
tarantool> print(tostring(err))
Invalid identifier 'Àaa' (expected printable symbols only or it is too long)
```

But if fact we're not restricted by correct indentifiers:

```lua
tarantool> _G['foo bar'] = function(...) print(...) end
tarantool> _G['1'] = function(...) print(...) end
tarantool> _G['end'] = function(...) print(...) end

tarantool> _G['foo bar'](1, 2, 3)
1	2	3
tarantool> _G['1'](1, 2, 3)
1	2	3
tarantool> _G['end'](1, 2, 3)
1	2	3

tarantool> box.cfg{listen = 3301}
tarantool> box.schema.user.grant('guest','read,write,execute,create,drop,alter','universe')
tarantool> net_box = require('net.box')
tarantool> c = net_box.connect(3301)

tarantool> c:call('foo bar', {1, 2, 3})
1	2	3
tarantool> c:call('1', {1, 2, 3})
1	2	3
tarantool> c:call('end', {1, 2, 3})
1	2	3
```

(`box.schema.func.create()` accepts such non-indentifier names too.)

Iproto calls and `box.func.<name>.call()` also accept a colon:

```
tarantool> _G.foo = {}
tarantool> _G.foo.bar = function(...) print(...) end
tarantool> box.schema.func.create('foo:bar')

tarantool> box.func['foo:bar']:call({1,2,3})
table: 0x40b19da0	1	2	3
tarantool> c:call('foo:bar', {1, 2, 3})
table: 0x40b19da0	1	2	3
```

(Locally we can even use non-string keys in `_G`.)

To be honest, I don't know what is better to do here. I don't like artificial restrictions. However I doubt that non identifier names are really used anywhere (as well as non ASCII ones).

Our purpose it to protect a user from an obvious mistake. What kind of errors we expect? Non-string value is likely a mistake. Non ASCII names too. Non identifier names as well.

So it looks practical to follow LuaJIT's rules (or even restrict them to ASCII):

1. Split an identifier by dots.
2. Verify those chunks.

At least, the chunks should NOT:

1. Be empty.
2. Be equal to a reserved keyword.
3. Contain control, space, punctuation symbols: we can grab LuaJIT's table
   from `lj_char.c` and use the `bit` module to reimplement `lj_char_isident`
   macro. Or use a Lua regexp for a chunk.
4. Start from a digit.

We can express the rules using a regexp (but not Lua regexp, see below). `lj_char_isident()` accepts the following ASCII characters: `[a-zA-Z0-9_]` (or `[%w_]`). So (if we restrict users to ASCII) the following regex would be correct: `^[%a_](%.[%w_]+)*$` if Lua would support repeating captures. It seems impossible to form a Lua regex that would accept 'foo.bar' like strings and will not accept 'foo..bar' for arbitrary number of chunks.

So we should actually split the identifier by dots and check the first and next chunks using regexps or using a symbol classes table (like `lj_char_bits`).

----

For the sake of completeness, there is a relevant issue in tarantool: https://github.com/tarantool/tarantool/issues/3405.

----

The code to check how `lj_char_isident()` and other macros work:

```
tarantool> lj_char_bits = {0, <...>} -- copy from lj_char.c
tarantool> function lj_char_isident(n)
               return bit.band(lj_char_bits[n + 2], 0x80) == 0x80
           end
tarantool> for i = 0, 255 do print(i, string.char(i), lj_char_isident(i)) end
```

ddl/check.lua
```diff
@@ -628,75 +628,147 @@ local function check_index(i, index, space)
...
+end
+
+local function check_custom_sharding_func(sharding_func_name)
+    local sharding_func = rawget(_G, sharding_func_name)
+    if sharding_func == nil then
+        return false
+    end
+
+    return is_callable(sharding_func)
+end
```
Totktonada, line 715: Why we use different rules for single component names and multiple dot separated component ones?

1. We accept any string for a single component name (say, starting from a digit). But not for components of multicomponent names.
2. We check whether a single component name is callable, but we don't do that for multicomponent one.

I don't thnk that there is any sense to divide names to 'custom' (single component) and 'module' (multiple component). A user can easily register a multicomponent name:

```lua
_G.foo = {}
function foo.bar(<...>)
    <...>
end
```

The single component name is the case of multicomponent one (with one component).

In fact, we know nothing about modules like vshard. We just perform `_G.<component 1>.<component 2>.<..and so on..>` to obtain a function.

ddl/check.lua
```diff
@@ -628,75 +628,151 @@ local function check_index(i, index, space)
...
+    -- because there is no easy way to get
+    -- metatable.__call of object with type `userdata`
+    if type(object) == 'userdata' then
+        return true
+    end
+
+    local object_metatable = getmetatable(object)
+    if type(object) == 'table' and object_metatable
+    and object_metatable.__call then
+        return type(object_metatable.__call) == 'function'
```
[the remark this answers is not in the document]
Totktonada: Thanks!

README.md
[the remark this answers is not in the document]
Totktonada: So technically we should add `boolean` into the list of Lua types.

ddl/check.lua
Totktonada: Okay. No, I don't think we should revisit functions visibility at whole within this task.

test/check_sharding_metadata_test.lua
```diff
@@ -219,3 +297,132 @@ function g.test_invalid_key_reference_path()
     )
     t.assert_equals(ok, nil)
 end
+
+function g.test_sharding_func_invalid_name()
+    g.space.sharding_func = 'bad_sharding_func'
```
Totktonada, line 302: Nit: The name is valid, but the function is not exists.

ddl/check.lua
Totktonada: My point is that 'vshard.router.bucket_id_strcrc32' is just a function in `_G` as any other. It seems, now it is implemented this way.

I would only drop the 'custom' term from the code: any callable is just callable, not 'known' or 'custom'. But I mentioned it in another comment, so I'll resolve this one.

README.md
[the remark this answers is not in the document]
Totktonada: I'll resolve it.

ddl/check.lua
Totktonada: We can detect a protected metatable (at least if `__metatable` is a string) and either accept it or decline. I would accept, because we can't say that it is not a callable (this way is consistent with our approach around cdata). But this should be rare case, so I'll not insist.

test/check_sharding_metadata_test.lua
AnaNek: Fixed.

ddl/check.lua
```diff
@@ -746,6 +822,49 @@ local function check_sharding_key(space)
...
+            )
+        end
+
+        if space.indexes.bucket_id ~= nil then
+            return nil, string.format(
+                "spaces[%q].indexes[%q]: bucket_id is used for sharding, " ..
+                "but there's no spaces[%q].sharding_key defined",
+                space.name, 'bucket_id', space.name
+            )
+        end
```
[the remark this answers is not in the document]
AnaNek: I guess this check is redundant.

Let's consider situation with no `sharding_key` / `sharding_func` in schema.
There are 4 cases:

1. No `bucket_id` field, no `bucket_id` index -> no error
2. `bucket_id` field exists, `bucket_id` index exists -> `spaces["space"].format["bucket_id"]: bucket_id is used for sharding, but there's no spaces["space"].sharding_key defined` error
3.  `bucket_id` field exists, no `bucket_id` index -> `spaces["space"].format["bucket_id"]: bucket_id is used for sharding, but there's no spaces["space"].sharding_key defined` error
4. No `bucket_id` field, `bucket_id` index exists -> `spaces["space"].indexes["bucket_id"].parts[1].path: path (bucket_id) referencing to unknown field"` error

So I suggest to remove this check, correct me if I am wrong.
(This check existed before this patch)

ddl/check.lua
```diff
@@ -628,75 +628,151 @@ local function check_index(i, index, space)
...
+        if type(space.sharding_func.body) ~= 'string' then
             return nil, string.format(
-                "spaces[%q].indexes[%q].parts[1].path: invalid field reference " ..
-                "(reference to bucket_id expected, got %s)",
-                space.name, 'bucket_id', space.indexes['bucket_id'].parts[1].path
+                "spaces[%q].sharding_func.body: bad value" ..
+                " (string expected, got %s)",
+                space.name, type(space.sharding_func.body)
             )
         end
```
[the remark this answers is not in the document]
AnaNek: Test was added:

```lua
function g.test_sharding_func_invalid_body()
    g.space.sharding_func = {body = 4}

    local ok, err = ddl.check_schema(g.schema)
    t.assert_equals(err,
        [[spaces["space"].sharding_func.body: bad value ]] ..
        [[(string expected, got number)]]
    )
    t.assert_equals(ok, nil)
end
```

README.md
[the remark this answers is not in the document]
AnaNek: Updated:

```diff
-key (number | table | string) – a sharding key.
+key (number | table | string | boolean) – a sharding key.
```

ddl/check.lua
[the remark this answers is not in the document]
ligurio: Screenshot from 2021-10-14 21-11-16

ddl/check.lua
AnaNek: Fixed:

```diff
-    and object_metatable and object_metatable.__call then
+    if (type(object) == 'table' or type(object) == 'userdata') then
+        -- if metatable type is not `table` -> metatable is protected ->
+        -- cannot detect metamethod `__call` exists
+        if object_metatable and type(object_metatable) ~= 'table' then
+            return true
+        end
+
         -- `__call` metamethod can be only the `function`
         -- and cannot be a `table` | `userdata` | `cdata`
         -- with `__call` methamethod on its own
-        return type(object_metatable.__call) == 'function'
+        if object_metatable and object_metatable.__call then
+            return type(object_metatable.__call) == 'function'
+        end
     end
```

AnaNek: Test for this case:

```lua
function g.test_user_sharding_func_protected_metatable()
    -- protected metatable has non-table type -> cannot detect
    -- object is callable-> ok
    local user_sharding_func_name = 'user_sharding_func'
    local user_sharding_func_table = setmetatable({}, {
        __call = function(_, key) return key end, __metatable = 'protected'
    })
    rawset(_G, user_sharding_func_name, user_sharding_func_table)
    g.space.sharding_func = user_sharding_func_name

    local ok, err = ddl.check_schema(g.schema)
    t.assert_equals(err, nil)
    t.assert_equals(ok, true)

    -- protected metatable has table type with __call metamethod -> cannot detect
    -- object is callable -> ok
    local user_sharding_func_name = 'user_sharding_func'
    local user_sharding_func_table = setmetatable({}, {
        __metatable = {__call = function(_, key) return key end}
    })
    rawset(_G, user_sharding_func_name, user_sharding_func_table)
    g.space.sharding_func = user_sharding_func_name

    local ok, err = ddl.check_schema(g.schema)
    t.assert_equals(err, nil)
    t.assert_equals(ok, true)
end
```

ddl/check.lua
```diff
@@ -628,75 +628,183 @@ local function check_index(i, index, space)
...
+        return false
+    end
+
+    if utils.LUA_KEYWORDS[name] then
+        return false
+    end
+
+    -- sharding func name must be sequence
+    -- of letters, digits, or underscore symbols,
+    -- and may not begin with a digit
```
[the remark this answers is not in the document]
Totktonada: I mean, that the identifier is not started from a digit.

ddl/check.lua
Totktonada: Regexp way would be simpler (if we're about a chunk validation), but I don't mind the LuaJIT way (it gains some UTF-8 support). If there will be other things to discuss around the dot syntax validation code, I'll add my questions as separate comments. I'm going to mark this thread as resolved.

Totktonada: Thanks for the update!

ddl/check.lua
Totktonada: Ouch, I see now. You're right. It is protected, but using a table in __metatable, so we cannot determine that it is protected.

ddl/check.lua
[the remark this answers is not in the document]
AnaNek: Sorry, my bad, I didn't call `rawset` in test case with digit as first symbol of sharding func name, that's why I got expected error but not because the check was correct the reason was incorrect test. Right now the test is fixed:

```lua
    -- first symbol is a digit
    local user_sharding_func_name = '5vshard.router.bucket_id_strcrc32'
    rawset(_G, user_sharding_func_name, function(key) return key end)
    g.space.sharding_func = user_sharding_func_name
    local ok, err = ddl.check_schema(g.schema)
    t.assert_equals(err,
        [[spaces["space"].sharding_func: unknown sharding_func "5vshard.router.bucket_id_strcrc32"]]
    )
    t.assert_equals(ok, nil)
    -- remove test data
    rawset(_G, user_sharding_func_name, nil)
```

Check is updated:

```lua
local function check_name_isident(name)
    if name == nil or name == '' then
        return false
    end

    -- sharding function name cannot
    -- be equal to lua keyword
    if utils.LUA_KEYWORDS[name] then
        return false
    end

    -- sharding function name cannot
    -- begin with a digit
    if tonumber(name:sub(1,1)) ~= nil then
        return false
    end

    -- sharding func name must be sequence
    -- of letters, digits, or underscore symbols
    for i = 1, #name do
        local char_number = string.byte(name:sub(i,i))
        if not utils.lj_char_isident(char_number) then
            return false
        end
    end

    return true
end
```

ddl/check.lua
```diff
@@ -746,6 +861,42 @@ local function check_sharding_key(space)
...
+        if space.fields.bucket_id ~= nil then
+            return nil, string.format(
+                "spaces[%q].format[%q]: bucket_id is used for sharding, " ..
+                "but there's no spaces[%q].sharding_key defined",
+                space.name, 'bucket_id', space.name
+            )
+        end
+
+        return true
+    end
```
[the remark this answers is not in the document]
AnaNek: https://github.com/tarantool/ddl/pull/72#discussion_r728882066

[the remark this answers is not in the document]
Totktonada: Sure, if we'll reference a non-existent field. But what if we'll reference some other one (existing)?

ddl/get.lua
```diff
@@ -52,12 +52,31 @@ local function _get_index(box_space, box_index)
...
+    if not record then
         return nil
     end
 
-    local record = box.space._ddl_sharding_key:get{space_name}
+    if record.sharding_func_body then
+        return {body = record.sharding_func_body}
+    end
+
+    return record.sharding_func_name
```
Totktonada, line 75: I would prefer explicit `~= nil` checks (for both fields), because accessing a nullable field in a middle of a tuple returns `box.NULL`. If we'll extend the space with more fields, this function may return `box.NULL` instead of `nil` and break some existing external code (if it doesn't check for `~= nil` as well).

ddl/check.lua
```diff
@@ -628,75 +628,190 @@ local function check_index(i, index, space)
...
+    -- be equal to lua keyword
+    if utils.LUA_KEYWORDS[name] then
+        return false
+    end
+
+    -- sharding function name cannot
+    -- begin with a digit
+    if tonumber(name:sub(1,1)) ~= nil then
+        return false
+    end
```
Totktonada, line 730: Nit: Since we already have the `lj_char_bits` table, we can define `lj_char_isdigit()`. Feel free to ignore, it is purely thing of taste (but looks logical for me, because we use `lj_char_isident()` just below).

ddl/check.lua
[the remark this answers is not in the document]
AnaNek: You are right, I hadn't thought about this case, this check was restored and a negative test for this error was added:

```lua
function g.test_invalid_index_bucket_id()
...
    -- replace reference field to non-bucket_id field in the `bucket_id` index
    local old_path = g.space.indexes[2].parts[1]
    table.remove(g.space.indexes[2].parts, 1)
    table.insert(g.space.indexes[2].parts, 1, {
        path = 'unsigned_nonnull', type = 'unsigned', is_nullable = false
    })

    local ok, err = ddl.check_schema(g.schema)
    t.assert_equals(err,
        [[spaces["space"].indexes["bucket_id"]: bucket_id is used for]] ..
        [[ sharding, but there's no spaces["space"].sharding_key defined]]
    )
    t.assert_equals(ok, nil)

    -- restore `bucket_id` field in space format
    table.insert(g.space.format, 1, old_format_field)
    -- restore reference field in `bucket_id` index
    table.insert(g.space.indexes[2].parts, 1, old_path)
...
end
```

ddl/get.lua
AnaNek: Updated:

```lua
local function get_sharding_func(space_name)
    local record = get_metadata(space_name, "sharding_func")
    if not record then
        return nil
    end

    if record.sharding_func_body ~= nil then
        return {body = record.sharding_func_body}
    end

    return record.sharding_func_name ~= nil and record.sharding_func_name
end
```

ddl/utils.lua
```diff
@@ -155,5 +162,6 @@ return {
     redundant_key = redundant_key,
     find_first_duplicate = find_first_duplicate,
     lj_char_isident = lj_char_isident,
+    lj_char_isdigit = lj_char_isdigit,
```
olegrok, line 165: Don't you consider using isdigit function from utf8 module? (https://www.tarantool.io/ru/doc/latest/reference/reference_lua/utf8/#utf8-isdigit)

Totktonada: No-no, let's do it consistently, please. I don't want one symbol check function from one source/lib and another from another one.

Totktonada (approved):
I don't see any problems anymore.

PR #78, opened 2021-09-24
Update info in README about _ddl_sharding_key space creation

AnaNek:
Space `_ddl_sharding_key` and it's format are a part of DDL API.
Users may use DDL spaces even no sharding keys or sharding spaces
are defined in DDL schema, so we need always create `_ddl_sharding_key`.
This behavior tested in `set_sharding_key.test_no_sharding_spaces`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,7 @@ in [documementation](https://www.tarantool.io/en/doc/latest/book/box/data_model/
 
     Return values: `true` if no error, otherwise return `nil, err`
 
-Additionally when a section `sharding_key` is not empty in YAML schema call of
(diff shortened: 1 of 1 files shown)


ligurio (changes requested):
Thanks for the patch!

I would add a couple of words with explanation to commit message.
Something like this:
"Space _ddl_sharding_key and it's format are a part of DDL API. Users may use DDL spaces even no sharding keys or sharding spaces are defined in DDL schema, so we need always create _ddl_sharding_key. This behavior tested in  set_sharding_key.test_no_sharding_spaces."

PR #80, opened 2021-09-27
Check sharding function (WIP)

ligurio:


ddl/check.lua
```diff
@@ -803,6 +804,21 @@ local function check_sharding_func(space)
                 space.name, type(space.sharding_func.body)
             )
         end
+
+        local sharding_key = {1, 2, 3}
```
Totktonada, line 808: What if I have the following sharding function?

```lua
function my_sharding_func(sharding_key)
    return my_hash(sharding_key[1], sharding_key[2]['foo']) % vshard.router.cfg.total_bucket_count + 1
end
```

Or, say, just check of correctness of the input data:

```lua
function my_sharding_func(sharding_key)
    checks({'number', 'string'})
    return my_hash(sharding_key) % vshard.router.cfg.total_bucket_count + 1
end
```

PR #77, opened 2021-09-06
Add bucket id calculation function

ligurio:
Closes #76
Fixes #83

PR #84, opened 2021-11-17
Switch to tagged version of GH action for source code checkout

ligurio:
It is better to stick to a latest tagged version, that considered as
stable, than to a current development version.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
             coveralls: true
     runs-on: [ubuntu-latest]
     steps:
-      - uses: actions/checkout@master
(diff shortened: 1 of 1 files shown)


Totktonada:
From CI:

> Run tarantoolctl rocks install luacov-coveralls 0.2.3-1 --server http://luarocks.org
>
> Error: Invalid argument: flag --server=<server> expects a parameter. See --help.
> Error: Process completed with exit code 1.

The change itself looks ok.

Totktonada:
Ah, I see, the fix will land together with PR #77.

ligurio:
> Error: Invalid argument: flag --server= expects a parameter. See --help.

Fixed in a separate commit.

PR #85, opened 2021-12-13
Add caching of sharding function

### LINKED ISSUE #82: Caching of sharding function
(There is more general task about caching #69, but I want to eat this pie piece by piece.)

We're going to add `ddl.bucket_id()` function (see #76). The function may be called quite frequently, so it worth to take care to its performance.

The `ddl.bucket_id()` function needs to know a sharding function. It is costly to obtain the function declaration / definition stored in the `_ddl_sharding_func` space, mainly due to those actions:

1. MsgPack decoding.
2. `loadstring()` is the function is declared as code (`{body = <...>}`).
3. Extra Lua GC pressure on re-cretion of usually same objects.

Ideally obtaining of the function should be just Lua table lookup. And it is possible to achieve.

The only way to track `_ddl_sharding_func` changes is to set a trigger on the space to track modifications ([on_replace](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/on_replace/))[^1]. Since it is not always possible to set a trigger when the module is just loaded[^2], I propose a trick.

The key idea is to generate an initial cache value and set the trigger when we access the sharding function information first time. After this the cache will be updated 'in background' (by the trigger) and we can just access the cache.

What to consider:

* Take care to access from different fibers (see the [locked()](https://github.com/tarantool/tarantool/blob/284797792771830215d87683e2796db830d47b07/src/box/lua/load_cfg.lua#L10-L22) function and the [synchronized](https://github.com/tarantool/synchronized/) module for examples).
* But implement this locking in a way that works good with hot reload (including cartridge's [one](https://github.com/tarantool/cartridge/blob/1d477838b097ccabd9ae7e0242951be12f4c8036/cartridge/hotreload.lua#L137-L142), which cleans up all globals).
* How to remove the old trigger if our code was unloaded (especially if _G was cleaned up by cartridge)? There are some ideas how to keep a state between reloads in https://github.com/tarantool/conf/issues/2 (at end of the issue description). A Lua/C module may also use Lua registry, but it is not our case.
* Don't use the cache in a transaction? That's slow, but how else we can ensure that the cache will correspond to the state visible in the transaction?
* Perform `loadstring()` when the function is defined as code (to don't do that each time).
* But don't save the function itself if it is defined by name: a user may replace it. Maybe just parse `'dot.notation'` into something like `{'dot', 'notation'}`.

Optimization trick:

We can use the trick with two implementations of the cache access function (see src/box/lua/load_cfg.lua in tarantool for example). The first function doing all the work: check whether the trigger is set (and the initial cache is generated), set the trigger and generate the cache if necessary, access cache, replace itself with the second function. The second function skips extra checks and just access the cache.

I'll note that the first function must not set the trigger unconditionally, because it may be called after hot reload. See https://github.com/tarantool/tarantool/issues/5826.

----

Looks a bit tricky, but doable. Opinions?

[^1]: I filed https://github.com/tarantool/tarantool/issues/6544 and https://github.com/tarantool/tarantool/issues/6545 to track it using the database schema version in a future. I think it may simplify some future code.
[^2]: box may be unconfigured (or not fully loaded), when the module is `require`d first time.

0x501D:
The ddl.bucket_id() function needs to know a sharding function.
It is costly to obtain the function declaration /
definition stored in the _ddl_sharding_func space.

Cache contains:
raw_tuple - raw sharding metadata, used for get_schema()
parsed_func_name - parsed dot notation (like {'foo', 'bar'})
callable - function ready to call, this offloads using of loadstring()
error - string with an error: not nil only if setting callable fails

Cache will be rebuilded if:
* _ddl_sharding_func space changed: cache sets _ddl_sharding_func:on_replace
  trigger
* schema changed: cache checks box.internal.schema_version changes

This patch does not serve hot reload techniques.
This entails an on_replace trigger duplication if hot reload occurs.
Hot reload support will be done in separate task:
https://github.com/tarantool/ddl/issues/87

Benchmarks - 10000000 bucket_id() calls (file test/bench_cache.lua):
Baseline (no DDL):     3.38s user 0.01s system 99% cpu 3.389 total
After this patch:
DDL with function body: 3.81s user 0.01s system 99% cpu 3.818 total
DDL with function name: 5.49s user 0.00s system 99% cpu 5.495 total
Before patch:
DDL with function body: 55.95s user 0.40s system 99% cpu 56.354 total
DDL with function name: 13.68s user 0.13s system 99% cpu 13.807 total

Closes #82

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,3 +66,27 @@ jobs:
       # Cleanup cached paths
       - run: tarantoolctl rocks remove cartridge
       - run: tarantoolctl rocks remove ddl
+  benchmark:
+    strategy:
+      fail-fast: false
+      matrix:
+        tarantool: ['1.10', '2.5', '2.6', '2.7']
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: tarantool/setup-tarantool@v1
+        with:
+          tarantool-version: ${{ matrix.tarantool }}
+
+      # Setup luatest
+      - name: Cache rocks
+        uses: actions/cache@v2
+        id: cache-rocks
+        with:
+          path: .rocks/
+          key: cache-rocks-${{ matrix.runs-on }}-bench-01
+
+      - run: tarantoolctl rocks install luatest
+        if: steps.cache-rocks.outputs.cache-hit != 'true'
+
+      - run: tarantool ./test/bench_cache.lua
--- a/README.md
+++ b/README.md
@@ -89,6 +89,11 @@ bucket identifier (number)
     - Calculate bucket id for a specified space and sharding key.
     Method uses sharding function specified in DDL schema.
 
+    Method is not transactional in the sense that it catches up
+    `_ddl_sharding_func` changes immediatelly: it may see changes that're
+    not committed yet and may see a state from another transaction,
+    which should not be visible in the current transaction.
+
     Return values: bucket_id if no error, otherwise return `nil, err`
 
 ## Input data format
--- a/ddl/cache.lua
+++ b/ddl/cache.lua
@@ -1,0 +1,124 @@
+local cache = nil
+
+local SPACE_NAME_IDX      = 1
+local SHARD_FUNC_NAME_IDX = 2
+local SHARD_FUNC_BODY_IDX = 3
+
+-- Build cache.
+--
+-- We don't need to call this function with any type of locking:
+-- _ddl_sharding_func is memtx space, so calling :pairs() on it
+-- is atomic
+--
+-- Cache structure format:
+--
+-- cache = {
+--     spaces = {
+--         [space_name] = {
+--             -- raw sharding metadata, used for ddl.get()
+--             raw_tuple = <tuple object> (<nil> at error),
+--             -- parsed dot notation (like {'foo', 'bar'})
+--             parsed_func_name = <table> or <nil>
+--             -- a function ready to call
+--             callable = <function> or <nil>,
+--             -- string with an error: not nil only if setting callable fails
+--             error = <string> or <nil>,
+--         }
+--     },
+--     -- current schema version
+--     schema_version = <...>,
+-- }
+
+-- function returns nothing
+local function cache_build()
+    -- clear cache
+    cache.spaces = {}
+
+    if box.space._ddl_sharding_func == nil then
+        return
+    end
+
+    for _, tuple in box.space._ddl_sharding_func:pairs() do
+        local space_name = tuple[SPACE_NAME_IDX]
+        local func_name = tuple[SHARD_FUNC_NAME_IDX]
+        local func_body = tuple[SHARD_FUNC_BODY_IDX]
+
+        cache.spaces[space_name] = {
+            raw_tuple = tuple
+        }
+
+        if func_body ~= nil then
+            local sharding_func, err = loadstring('return ' .. func_body)
+            if sharding_func == nil then
+                cache.spaces[space_name].error =
+                    string.format("Body is incorrect in sharding_func for space (%s): %s",
+                    space_name, err)
+            else
+                cache.spaces[space_name].callable = sharding_func()
+            end
+        elseif func_name ~= nil then
+            -- we cannot save the function itself into the cache,
+            -- because the function can be changed in runtime and
+            -- there is no way to catch this change
+            local chunks = string.split(func_name, '.')
+            cache.spaces[space_name].parsed_func_name = chunks
+        end
+    end
+
+    cache.schema_version = box.internal.schema_version()
+
+end
+
+-- Rebuild cache if _ddl_sharding_func space changed.
+local function cache_set_trigger()
+    if box.space._ddl_sharding_func == nil then
+        return
+    end
+
+    local trigger_found = false
+
+    for _, func in pairs(box.space._ddl_sharding_func:on_replace()) do
+        if func == cache_build then
+            trigger_found = true
+            break
+        end
+    end
+
+    if not trigger_found then
+        box.space._ddl_sharding_func:on_replace(cache_build)
+    end
+end
+
+-- Get data from cache.
+-- Returns all cached data for "space_name" or nil.
+local function cache_get(space_name)
+    if space_name == nil then
+        return nil
+    end
+
+    -- using tarantool internal API.
+    -- this is not reliable, but it is the only way to track
+    -- schema_version changes. Fix it if a public method appears:
+    -- https://github.com/tarantool/tarantool/issues/6544
+    local schema_version = box.internal.schema_version()
+
+    if not cache then
+        cache = {}
+        cache_build()
+        cache_set_trigger()
+    end
+
+    -- rebuild cache if database schema changed
+    if schema_version ~= cache.schema_version then
+        cache_build()
+        cache_set_trigger()
+    end
+
+    return cache.spaces[space_name]
+end
+
+return {
+    internal = {
+        get = cache_get,
+    }
+}
--- a/ddl/get.lua
+++ b/ddl/get.lua
@@ -1,4 +1,5 @@
 local utils = require('ddl.utils')
+local cache = require('ddl.cache')
 local ddl_check = require('ddl.check')
 
 local function _get_index_field_path(space, index_part)
@@ -66,10 +67,17 @@ local function get_metadata(space_name, metadata_name)
 end
 
 local function get_sharding_func(space_name)
-    local record = get_metadata(space_name, "sharding_func")
-    if not record then
+    return cache.internal.get(space_name)
+end
+
+local function get_sharding_func_raw(space_name)
+    local record = cache.internal.get(space_name)
+
+    if not record or not record.raw_tuple then
         return nil
     end
+
+    record = record.raw_tuple
 
     if record.sharding_func_body ~= nil then
         return {body = record.sharding_func_body}
@@ -97,7 +105,7 @@ local function get_space_schema(space_name)
     space_ddl.engine = box_space.engine
     space_ddl.format = box_space:format()
     space_ddl.sharding_key = get_sharding_key(space_name)
-    space_ddl.sharding_func = get_sharding_func(space_name)
+    space_ddl.sharding_func = get_sharding_func_raw(space_name)
     for _, field in ipairs(space_ddl.format) do
         if field.is_nullable == nil then
             field.is_nullable = false
@@ -115,21 +123,20 @@ local function get_space_schema(space_name)
 end
 
 local function prepare_sharding_func_for_call(space_name, sharding_func_def)
-    if type(sharding_func_def) == 'string' then
-        local sharding_func = utils.get_G_function(sharding_func_def)
+    if sharding_func_def.error ~= nil then
+        return nil, sharding_func_def.error
+    end
+
+    if sharding_func_def.parsed_func_name ~= nil then
+        local sharding_func = utils.get_G_function(sharding_func_def.parsed_func_name)
         if sharding_func ~= nil and
            ddl_check.internal.is_callable(sharding_func) == true then
             return sharding_func
         end
     end
 
-    if type(sharding_func_def) == 'table' then
-        local sharding_func, err = loadstring('return ' .. sharding_func_def.body)
-        if sharding_func == nil then
-            return nil, string.format(
-                "Body is incorrect in sharding_func for space (%s): %s", space_name, err)
-        end
-        return sharding_func()
+    if sharding_func_def.callable ~= nil then
+        return sharding_func_def.callable
     end
 
     return nil, string.format(
--- a/ddl/utils.lua
+++ b/ddl/utils.lua
@@ -189,9 +189,19 @@ end
 -- split sharding func name in dot notation by dot
 -- foo.bar.baz -> chunks: foo bar baz
 -- foo -> chunks: foo
+--
+-- func_name parameter may be a string in dot notation or table
+-- if func_name type is of type table it is assumed that it is already split
 local function get_G_function(func_name)
-    local chunks = string.split(func_name, '.')
     local sharding_func = _G
+    local chunks
+
+    if type(func_name) == 'string' then
+        chunks = string.split(func_name, '.')
+    else
+        chunks = func_name
+    end
+
     -- check is the each chunk an identifier
     for _, chunk in pairs(chunks) do
         if not check_name_isident(chunk) or sharding_func == nil then
--- a/test/bench_cache.lua
+++ b/test/bench_cache.lua
@@ -1,0 +1,140 @@
+#!/usr/bin/env tarantool
+
+local ddl = require('ddl')
+local db = require('test.db')
+local clock = require('clock')
+local helper = require('test.helper')
+
+local ITERS = 10000000
+
+local primary_index = {
+    type = 'HASH',
+    unique = true,
+    parts = {
+        {path = 'string_nonnull', is_nullable = false, type = 'string'},
+        {path = 'unsigned_nonnull', is_nullable = false, type = 'unsigned'},
+    },
+    name = 'primary'
+}
+
+local bucket_id_idx = {
+    type = 'TREE',
+    unique = false,
+    parts = {{path = 'bucket_id', type = 'unsigned', is_nullable = false}},
+    name = 'bucket_id'
+}
+
+local function space_init()
+    db.drop_all()
+
+    local space = {
+        engine = 'memtx',
+        is_local = true,
+        temporary = false,
+        format = table.deepcopy(helper.test_space_format())
+    }
+    table.insert(space.format, 1, {
+        name = 'bucket_id', type = 'unsigned', is_nullable = false
+    })
+
+    space.indexes = {
+        table.deepcopy(primary_index),
+        table.deepcopy(bucket_id_idx)
+    }
+    space.sharding_key = {'unsigned_nonnull', 'integer_nonnull'}
+    local schema = {
+        spaces = {
+            space = space,
+        }
+    }
+
+    return schema
+end
+
+local function run_body()
+    local schema = space_init()
+    schema.spaces.space.sharding_func = {
+        body = helper.sharding_func_body
+    }
+
+    local _, err = ddl.set_schema(schema)
+    if err then
+        print(err)
+        os.exit()
+    end
+
+    for i=1,ITERS do
+        local _, err = ddl.bucket_id('space', i)
+        if err then
+            print(err)
+            os.exit()
+        end
+    end
+end
+
+local function run_baseline()
+    local schema = space_init()
+    schema.spaces.space.sharding_func = {
+        body = helper.sharding_func_body
+    }
+
+    local _, err = ddl.set_schema(schema)
+    if err then
+        print(err)
+        os.exit()
+    end
+
+    for i=1,ITERS do
+        helper.sharding_func(i)
+        if err then
+            print(err)
+            os.exit()
+        end
+    end
+end
+
+local function run_name()
+    local schema = space_init()
+    local sharding_func_name = 'sharding_func'
+    rawset(_G, sharding_func_name, helper.sharding_func)
+    schema.spaces.space.sharding_func = sharding_func_name
+
+    local _, err = ddl.set_schema(schema)
+    if err then
+        print(err)
+        os.exit()
+    end
+
+    for i=1,ITERS do
+        local _, err = ddl.bucket_id('space', i)
+        if err then
+            print(err)
+            os.exit()
+        end
+    end
+end
+
+local function main()
+    local benchs = {
+        {"baseline",      run_baseline},
+        {"function body", run_body    },
+        {"function name", run_name    }
+    }
+
+    -- suppress logs from tarantool
+    require("log").level(1)
+    db.init()
+
+    for _,b in pairs(benchs) do
+        io.write(string.format("* cache benchmark: %s... ", b[1]))
+        io.flush()
+        local start = clock.monotonic()
+        b[2]()
+        local finish = clock.monotonic()
+        print(string.format("%0.3f seconds", finish - start))
+    end
+
+    os.exit()
+end
+
+main()
--- a/test/cache_test.lua
+++ b/test/cache_test.lua
@@ -1,0 +1,247 @@
+#!/usr/bin/env tarantool
+
+local t = require('luatest')
+local db = require('test.db')
+local ddl = require('ddl')
+local cache = require('ddl.cache')
+local helper = require('test.helper')
+
+local SPACE_NAME_IDX      = 1
+local SHARD_FUNC_NAME_IDX = 2
+local SHARD_FUNC_BODY_IDX = 3
+
+local primary_index = {
+    type = 'HASH',
+    unique = true,
+    parts = {
+        {path = 'string_nonnull', is_nullable = false, type = 'string'},
+        {path = 'unsigned_nonnull', is_nullable = false, type = 'unsigned'},
+    },
+    name = 'primary'
+}
+
+local bucket_id_idx = {
+    type = 'TREE',
+    unique = false,
+    parts = {{path = 'bucket_id', type = 'unsigned', is_nullable = false}},
+    name = 'bucket_id'
+}
+
+local func_body_first  = 'function(key) return key + 1 end'
+local func_body_second = 'function(key) return key + 2 end'
+
+local function space_init(g)
+    db.drop_all()
+
+    g.space = {
+        engine = 'memtx',
+        is_local = true,
+        temporary = false,
+        format = table.deepcopy(helper.test_space_format())
+    }
+    table.insert(g.space.format, 1, {
+        name = 'bucket_id', type = 'unsigned', is_nullable = false
+    })
+
+    g.space.indexes = {
+        table.deepcopy(primary_index),
+        table.deepcopy(bucket_id_idx)
+    }
+    g.space.sharding_key = {'unsigned_nonnull', 'integer_nonnull'}
+    g.schema = {
+        spaces = {
+            space = g.space,
+        }
+    }
+end
+
+local g = t.group()
+g.before_all(db.init)
+g.before_each(space_init)
+
+function g.test_cache_processed_func_body()
+    g.schema.spaces.space.sharding_func = {
+        body = func_body_first
+    }
+    local ok, err = ddl.set_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(ok, true)
+
+    local res = cache.internal.get('space')
+    t.assert(res)
+    t.assert(res.callable)
+    t.assert_equals(res.error, nil)
+    t.assert_type(res.callable, 'function')
+    t.assert_equals(res.callable(42), 43)
+    local id, err = ddl.bucket_id('space', 42)
+    t.assert_equals(err, nil)
+    t.assert_type(id, 'number')
+    t.assert_equals(id, 43)
+end
+
+function g.test_cache_processed_func_name()
+    local sharding_func_name = 'sharding_func'
+    rawset(_G, sharding_func_name, function(key) return key + 1 end)
+    g.schema.spaces.space.sharding_func = sharding_func_name
+
+    local ok, err = ddl.set_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(ok, true)
+
+    local res = cache.internal.get('space')
+    t.assert(res)
+    t.assert(res.parsed_func_name)
+    t.assert_type(res.parsed_func_name, 'table')
+    t.assert_equals(res.parsed_func_name[1], 'sharding_func')
+    local id, err = ddl.bucket_id('space', 42)
+    t.assert_equals(err, nil)
+    t.assert_type(id, 'number')
+    t.assert_equals(id, 43)
+
+    rawset(_G, sharding_func_name, nil)
+end
+
+function g.test_cache_nil_space()
+    local sharding_func_name = 'sharding_func'
+    rawset(_G, sharding_func_name, function(key) return key + 1 end)
+    g.schema.spaces.space.sharding_func = sharding_func_name
+
+    local ok, err = ddl.set_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(ok, true)
+
+    local res = cache.internal.get()
+    t.assert_equals(res, nil)
+
+    rawset(_G, sharding_func_name, nil)
+end
+
+-- when caching function name, we must store in cache
+-- only name, because body can be changed manually
+function g.test_body_changed_manually()
+    local sharding_func_name = 'sharding_func'
+    rawset(_G, sharding_func_name, function(key) return key + 10 end)
+    g.schema.spaces.space.sharding_func = sharding_func_name
+
+    local ok, err = ddl.set_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(ok, true)
+
+    local id, err = ddl.bucket_id('space', 42)
+    t.assert_equals(err, nil)
+    t.assert_type(id, 'number')
+    t.assert_equals(id, 52)
+
+    rawset(_G, sharding_func_name, function(key) return key + 11 end)
+    local id, err = ddl.bucket_id('space', 42)
+    t.assert_equals(err, nil)
+    t.assert_type(id, 'number')
+    t.assert_equals(id, 53)
+
+    rawset(_G, sharding_func_name, nil)
+end
+
+function g.test_cache_schema_changed()
+    g.schema.spaces.space.sharding_func = {
+        body = func_body_first
+    }
+    local ok, err = ddl.set_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(ok, true)
+
+    local res = cache.internal.get('space')
+    t.assert(res)
+    t.assert(res.raw_tuple)
+    t.assert_equals(res.raw_tuple[SPACE_NAME_IDX], 'space')
+    t.assert_equals(res.raw_tuple[SHARD_FUNC_NAME_IDX], nil)
+    t.assert_equals(res.raw_tuple[SHARD_FUNC_BODY_IDX], func_body_first)
+
+    space_init(g)
+
+    local res = cache.internal.get('space')
+    t.assert_equals(res, nil)
+    local id, err = ddl.bucket_id('space', 42)
+    t.assert_equals(err, 'No sharding function specified in DDL schema of space (space)')
+    t.assert_equals(id, nil)
+end
+
+-- test for _ddl_sharding_func:on_replace() trigger
+-- cache must be rebuilded on space changed/updated
+function g.test_cache_space_updated_body()
+    g.schema.spaces.space.sharding_func = {
+        body = func_body_first
+    }
+    local ok, err = ddl.set_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(ok, true)
+
+    local res = cache.internal.get('space')
+    t.assert(res)
+    t.assert(res.callable)
+    t.assert_type(res.callable, 'function')
+    t.assert_equals(res.callable(42), 43)
+    t.assert(res.raw_tuple)
+    t.assert_equals(res.error, nil)
+    t.assert_equals(res.raw_tuple[SPACE_NAME_IDX], 'space')
+    t.assert_equals(res.raw_tuple[SHARD_FUNC_NAME_IDX], nil)
+    t.assert_equals(res.raw_tuple[SHARD_FUNC_BODY_IDX], func_body_first)
+    local id, err = ddl.bucket_id('space', 42)
+    t.assert_equals(err, nil)
+    t.assert_type(id, 'number')
+    t.assert_equals(id, 43)
+
+    box.space._ddl_sharding_func
+       :update({'space'}, {{'=', SHARD_FUNC_BODY_IDX, func_body_second}})
+
+    local res = cache.internal.get('space')
+    t.assert(res)
+    t.assert(res.callable)
+    t.assert_type(res.callable, 'function')
+    t.assert_equals(res.callable(42), 44)
+    t.assert(res.raw_tuple)
+    t.assert_equals(res.error, nil)
+    t.assert_equals(res.raw_tuple[SPACE_NAME_IDX], 'space')
+    t.assert_equals(res.raw_tuple[SHARD_FUNC_NAME_IDX], nil)
+    t.assert_equals(res.raw_tuple[SHARD_FUNC_BODY_IDX], func_body_second)
+    local id, err = ddl.bucket_id('space', 42)
+    t.assert_equals(err, nil)
+    t.assert_type(id, 'number')
+    t.assert_equals(id, 44)
+end
+
+function g.test_cache_space_updated_name()
+    local sharding_func_first = 'sharding_func_first'
+    rawset(_G, sharding_func_first, function(key) return key + 1 end)
+    rawset(_G, 'sharding_func_second', function(key) return key + 2 end)
+    g.schema.spaces.space.sharding_func = sharding_func_first
+    local ok, err = ddl.set_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(ok, true)
+
+    local res = cache.internal.get('space')
+    t.assert(res)
+    t.assert(res.parsed_func_name)
+    t.assert_type(res.parsed_func_name, 'table')
+    t.assert_equals(res.parsed_func_name[1], 'sharding_func_first')
+    t.assert_equals(res.raw_tuple[SPACE_NAME_IDX], 'space')
+    t.assert_equals(res.raw_tuple[SHARD_FUNC_NAME_IDX], 'sharding_func_first')
+    local id, err = ddl.bucket_id('space', 42)
+    t.assert_equals(err, nil)
+    t.assert_type(id, 'number')
+    t.assert_equals(id, 43)
+
+    box.space._ddl_sharding_func
+       :update({'space'}, {{'=', SHARD_FUNC_NAME_IDX, 'sharding_func_second'}})
+
+    local res = cache.internal.get('space')
+    t.assert(res)
+    t.assert(res.parsed_func_name)
+    t.assert_type(res.parsed_func_name, 'table')
+    t.assert_equals(res.parsed_func_name[1], 'sharding_func_second')
+    t.assert_equals(res.raw_tuple[SPACE_NAME_IDX], 'space')
+    t.assert_equals(res.raw_tuple[SHARD_FUNC_NAME_IDX], 'sharding_func_second')
+    local id, err = ddl.bucket_id('space', 42)
+    t.assert_equals(err, nil)
+    t.assert_type(id, 'number')
+    t.assert_equals(id, 44)
+end


.luacheckrc
```diff
@@ -2,3 +2,6 @@ redefined = false
 include_files = {"**/*.lua", "*.rockspec", "*.luacheckrc"}
 exclude_files = {".rocks/", "tmp/"}
 max_line_length = 120
+ignore = {
+    "113", "143"
+}
```
0x501D, line 7: Вот это залетело случайно, уберу. Но оно соответствует тестам

ddl/cache.lua
```diff
@@ -0,0 +1,99 @@
+local fiber = require('fiber')
+
+local cache = {
+    spaces = {},
+    _initialised = false,
+    _triggers = {},
```
0x501D, line 6: _triggers will be removed in next commit

ddl/cache.lua
```diff
@@ -0,0 +1,110 @@
...
+
+    if not trigger_found then
+        box.space._ddl_sharding_func:on_replace(cache_build)
+    end
+
+    cache._initialised = true
+    cache._lock:get()
+end
+
+local function cache_get(space_name, entry)
```
ligurio, line 73: It is not clear what `entry` means. Please put a comment for function and describe there arguments and returned values.

ddl/cache.lua
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    if box.space._ddl_sharding_func == nil then
+        cache._lock:get()
+        return nil
+    end
+
+    for _, tuple in box.space._ddl_sharding_func:pairs() do
+        cache.spaces[tuple[SPACE_NAME_IDX]] = {}
+        cache.spaces[tuple[SPACE_NAME_IDX]]['sharding_func'] = {}
```
ligurio, line 36: can be replaced to:
```lua
cache.spaces[tuple[SPACE_NAME_IDX]]['sharding_func'] = {
    raw = tuple
}
```

ddl/cache.lua
```diff
@@ -0,0 +1,99 @@
+local fiber = require('fiber')
+
+local cache = {
+    spaces = {},
```
ligurio, line 4: I think it would be useful to describe a structure of cache for every record in comment or as a table with `nil` as values.
For example:
```lua
space = {
    'space_name' = {
        'raw' = nil, -- raw sharding metadata, used for ddl.get().
        'processed' = nil -- sharding function that is ready to call or a string with error.
    }
}
```

ligurio (changes requested):
Thanks for the patch! See my comments inline.

Also please add a 'Closes #82' to commit message and a number of issue to PR description.

ddl/cache.lua
```diff
@@ -0,0 +1,110 @@
+local log = require('log')
+local fiber = require('fiber')
+
+local cache = { 
+    spaces = {}, 
+    _initialised = false,
+    _triggers = {},
+    _lock = fiber.channel(1)
```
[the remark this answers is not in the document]
0x501D: fixed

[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,110 @@
+local log = require('log')
+local fiber = require('fiber')
+
+local cache = { 
+    spaces = {}, 
+    _initialised = false,
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,110 @@
...
+local function cache_build()
+    -- Prevent cache_build() from being called concurrently by different fibers.
+    local unlocked = cache._lock:put(true, CACHE_LOCK_TIMEOUT)
+    local trigger_found = false
+
+    if not unlocked then
+        return
+    end
+
+    -- clear cache
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,110 @@
...
+    for key, _ in pairs(cache.spaces) do
+        cache.spaces[key] = nil
+    end
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,110 @@
...
+    if entry ~= nil then
+        return (cache.spaces[space_name] ~= nil) and
+            cache.spaces[space_name][entry] or nil
+    end
+
+    return cache.spaces[space_name]
+end
+
+return {
+    get = cache_get,
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,110 @@
...
+    if box.space._ddl_sharding_func == nil then
+        cache._lock:get()
+        return nil
+    end
+
+    for _, tuple in box.space._ddl_sharding_func:pairs() do
+        cache.spaces[tuple[SPACE_NAME_IDX]] = {}
+        cache.spaces[tuple[SPACE_NAME_IDX]]['sharding_func'] = {}
+
+        cache.spaces[tuple[SPACE_NAME_IDX]]['sharding_func']['raw'] = tuple
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
0x501D: fixed

ddl/cache.lua
0x501D: fixed

ddl/cache.lua
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,108 @@
...
+--   fetch_lock -- locking based on fiber.channel()
+-- }
+local function cache_build()
+    -- prevent cache_build() from being called concurrently by different fibers
+    local unlocked = cache.fetch_lock:put(true, CACHE_LOCK_TIMEOUT)
+    local trigger_found = false
+
+    if not unlocked then
+        return
+    end
```
Totktonada, line 29: Something going wrong... shouldn't we return an error?

ddl/cache.lua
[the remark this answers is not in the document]
Totktonada: 'Fetch' is usually about something transmitted over a network. Here we need the lock to don't set a trigger twice, while reading from a memtx space could be implemented without locking at all. I'm not sure we can set triggers in a transaction, but channels solution should work.

To sum up: it is locking for setting triggers, not for network transmittion. Either just `lock` or `cache_build_lock` would fit better (to be honest, I would prefer the shorter).

ddl/utils.lua
```diff
@@ -190,8 +190,16 @@ end
 -- foo.bar.baz -> chunks: foo bar baz
 -- foo -> chunks: foo
 local function get_G_function(func_name)
-    local chunks = string.split(func_name, '.')
     local sharding_func = _G
+    local chunks
+
+    -- function name received from cache is already splitted
```
Totktonada, line 196: Please, add it to the functon contract (the comment prior to the function). When `func_name` was always a string, it was obvious. Now it worth to highlight that the parameter may be a string in dot notation or table (list of keys that represents the same).

ddl/cache.lua
```diff
@@ -0,0 +1,108 @@
...
+local SHARD_FUNC_BODY_IDX = 3
+
+-- Build cache and setup "on_replace" trigger.
+--
+-- Cache structure format:
+-- cache = {
+--   spaces = {
+--     'space_name' = {
+--       raw = {},      -- raw sharding metadata, used for ddl.get()
+--       processed = {} -- sharding function that is ready to call or a string with error
```
Totktonada, line 17: It is a `table` with parsed dot notation (like `{'foo', 'bar'}`) or a `function` ready to call (or a `string` with an error). It is not obvious that 'sharding function' words assume the former two (table or function).

ddl/cache.lua
```diff
@@ -0,0 +1,110 @@
...
+    local metadata_space_name = string.format("_ddl_%s", metadata_name)
+
+    if box.space[metadata_space_name] == nil then
+        return nil
+    end
+
+    return box.space[metadata_space_name]:get{space_name}
+end
+
+local function cache_build()
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,110 @@
...
+        cache.spaces[key] = nil
+    end
+
+    if box.space._ddl_sharding_func == nil then
+        cache._lock:get()
+        return nil
+    end
+
+    for _, tuple in box.space._ddl_sharding_func:pairs() do
+        cache.spaces[tuple[SPACE_NAME_IDX]] = {}
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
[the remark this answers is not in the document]
0x501D: renamed to "lock"

ddl/cache.lua
0x501D: fixed

ddl/utils.lua
0x501D: fixed

ddl/cache.lua
0x501D: fixed

ligurio (changes requested):

ddl/cache.lua
```diff
@@ -0,0 +1,128 @@
...
+        return
+    end
+
+    for _, tuple in box.space._ddl_sharding_func:pairs() do
+        local space_name = tuple[SPACE_NAME_IDX]
+        local func_name = tuple[SHARD_FUNC_NAME_IDX]
+        local func_body = tuple[SHARD_FUNC_BODY_IDX]
+
+        cache.spaces[space_name] = {}
+        cache.spaces[space_name].sharding_func = { raw = tuple }
```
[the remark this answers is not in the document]
0x501D: This is done for future cache extension. I can remove it.

ddl/cache.lua
```diff
@@ -0,0 +1,128 @@
...
+                cache.spaces[space_name].sharding_func.processed =
+                    sharding_func()
+            end
+        elseif func_name ~= nil then
+            local chunks = string.split(func_name, '.')
+            cache.spaces[space_name].sharding_func.processed = chunks
+        end
+    end
+
+    cache._schema_version = box.internal.schema_version()
```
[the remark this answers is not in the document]
0x501D: No, but it is the only way to get this data. We have an issue for getting public function: https://github.com/tarantool/tarantool/issues/6544

ddl/cache.lua
```diff
@@ -0,0 +1,128 @@
...
+
+        if not status or err ~= nil then
+            return err
+        end
+    end
+end
+
+-- Build cache and setup "on_replace" trigger.
+--
+-- Cache structure format:
```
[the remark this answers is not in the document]
0x501D: cache_build returns nothing

ddl/cache.lua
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,128 @@
...
+    end
+
+    cache._schema_version = box.internal.schema_version()
+end)
+
+-- Rebuild cache if _ddl_sharding_func space changed.
+local function cache_set_trigger()
+    if box.space._ddl_sharding_func == nil then
+        return
+    end
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,128 @@
...
+end
+
+-- Get data from cache.
+-- Returns all cached data for "space_name".
+local function cache_get(space_name)
+    if space_name == nil then
+        return nil
+    end
+
+    local schema_version = box.internal.schema_version()
```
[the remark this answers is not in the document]
0x501D: the same as I mentioned above

ddl/cache.lua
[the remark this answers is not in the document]
ligurio: Could you add a link to the issue and explanation regarding using internal API to commit message?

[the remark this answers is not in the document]
0x501D: Sure

ligurio (changes requested):

ddl/cache.lua
```diff
@@ -0,0 +1,127 @@
...
+                    string.format("Body is incorrect in sharding_func for space (%s): %s",
+                    space_name, err)
+            else
+                cache.spaces[space_name].processed =
+                    sharding_func()
+            end
+        elseif func_name ~= nil then
+            local chunks = string.split(func_name, '.')
+            cache.spaces[space_name].processed = chunks
+        end
```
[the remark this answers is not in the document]
0x501D: bicket_id() will return "No sharding function specified in DDL schema of space (%s)" as exppected.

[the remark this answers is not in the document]
0x501D: description added

ddl/cache.lua
```diff
@@ -0,0 +1,127 @@
...
+                cache.spaces[space_name].processed =
+                    string.format("Body is incorrect in sharding_func for space (%s): %s",
+                    space_name, err)
+            else
+                cache.spaces[space_name].processed =
+                    sharding_func()
+            end
+        elseif func_name ~= nil then
+            local chunks = string.split(func_name, '.')
+            cache.spaces[space_name].processed = chunks
```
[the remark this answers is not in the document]
0x501D: split from get_G_function was offloaded to cache. That's all.

ddl/cache.lua
```diff
@@ -0,0 +1,127 @@
...
+        return
+    end
+
+    for _, tuple in box.space._ddl_sharding_func:pairs() do
+        local space_name = tuple[SPACE_NAME_IDX]
+        local func_name = tuple[SHARD_FUNC_NAME_IDX]
+        local func_body = tuple[SHARD_FUNC_BODY_IDX]
+
+        cache.spaces[space_name] = {}
+        cache.spaces[space_name].raw = tuple
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,127 @@
...
+                cache.spaces[space_name].processed =
+                    sharding_func()
+            end
+        elseif func_name ~= nil then
+            local chunks = string.split(func_name, '.')
+            cache.spaces[space_name].processed = chunks
+        end
+    end
+
+    cache._schema_version = box.internal.schema_version()
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,127 @@
...
+        end
+    end
+
+    if not trigger_found then
+        box.space._ddl_sharding_func:on_replace(cache_build)
+    end
+end
+
+-- Get data from cache.
+-- Returns all cached data for "space_name".
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,127 @@
...
+        if not status or err ~= nil then
+            return err
+        end
+    end
+end
+
+-- Build cache.
+--
+-- Cache structure format:
+-- cache = {
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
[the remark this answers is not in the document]
0x501D: fixed

ddl/get.lua
```diff
@@ -66,11 +67,24 @@ local function get_metadata(space_name, metadata_name)
 end
 
 local function get_sharding_func(space_name)
-    local record = get_metadata(space_name, "sharding_func")
```
ligurio, line 69: We have a function `get_metadata()` used in function `get_sharding_key()` and `get_sharding_func()`. After your patch, it will be used in `get_sharding_key()` only. I remember that you implement a cache for sharding *function*. But it is a bit inconsistent: `get_sharding_func` uses a cache with sharding metadata and while `get_sharding_key` does not.

Perhaps we need to replace `get_metadata()` with `cache.internal.get` everywhere and remove `get_metadata` at all.

ddl/get.lua
```diff
@@ -115,21 +129,21 @@ local function get_space_schema(space_name)
...
-        end
-        return sharding_func()
+    if type(sharding_func_def) == 'function' then
+        return sharding_func_def
+    end
+
+    -- error from cache
+    if type(sharding_func_def) == 'string' then
+        return nil, sharding_func_def
     end
```
ligurio, line 147: We can remove this function at all and put already prepared sharding function to the cache. What do you think?

test/cache_test.lua
```diff
@@ -0,0 +1,159 @@
...
+
+local bucket_id_idx = {
+    type = 'TREE',
+    unique = false,
+    parts = {{path = 'bucket_id', type = 'unsigned', is_nullable = false}},
+    name = 'bucket_id'
+}
+
+local func_body_first  = 'function() return 42 end'
+local func_body_second = 'function() return 24 end'
```
ligurio, line 31: What is a difference for `func_body_first` and `func_body_second` from testing point of view?

test/cache_test.lua
```diff
@@ -0,0 +1,159 @@
...
+            space = g.space,
+        }
+    }
+end
+
+local g = t.group()
+g.before_all(db.init)
+g.before_each(function()
+    rebuild_db(g)
+end)
```
ligurio, line 62: Usually test setup used for initialization and other steps required for test and teardown used to cleanup test data or destroy test data. You are using  both in `rebuild_db` and call it in `before_each`, function for test setup. What is a point of this?

Can we do like this:

```lua

local function space_init(g)
    g.space = {
        engine = 'memtx',
        is_local = true,
        temporary = false,
        format = table.deepcopy(helper.test_space_format())
    }
    table.insert(g.space.format, 1, {
        name = 'bucket_id', type = 'unsigned', is_nullable = false
    })

    g.space.indexes = {
        table.deepcopy(primary_index),
        table.deepcopy(bucket_id_idx)
    }
    g.space.sharding_key = {'unsigned_nonnull', 'integer_nonnull'}
    g.schema = {
        spaces = {
            space = g.space,
        }
    }
end

g.before_each() = space_init()

g.after_each = db.drop_all()
```

With such setup and teardown, it is clear what  happen before and after each test.

test/cache_test.lua
```diff
@@ -0,0 +1,159 @@
...
+    }
+    local ok, err = ddl.set_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(ok, true)
+
+    local res = cache.internal.get('space')
+    t.assert(res)
+    t.assert(res.processed)
+    res = res.processed
+    t.assert(type(res) == 'function')
```
ligurio, line 76: Why do you need to create a shortcut for `res.processed` in `res` variable and cannot use it directly in assert?

ligurio (changes requested):

test/cache_test.lua
0x501D: get different results. so we can be sure that trigger was called and cache is actual

ddl/get.lua
0x501D: We cannot because sharding function, defined by name, may not be exists at runtime (e.g deleted by user). So we need this checks: utils.get_G_function( ...

test/cache_test.lua
0x501D: Thanks for the tip! I will use this method.

ddl/get.lua
Totktonada: Your point is valid, but I'm afraid that if we'll going to implement caching of the sharding keys here (or even stubs for it), we'll expand the scope of the task too much.

test/cache_test.lua
0x501D: fixed

test/cache_test.lua
0x501D: fixed

ddl/cache.lua
```diff
@@ -0,0 +1,136 @@
...
+--   spaces = {
+--     'space_name' = {
+--       raw = {},      -- raw sharding metadata, used for ddl.get()
+--       processed = {} -- table with parsed dot notation (like {'foo', 'bar'})
```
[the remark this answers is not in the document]
0x501D: fixed

test/cache_test.lua
0x501D: --g.after_each = db.drop_all()
we need to clear db before first test call also, db is dirty after other tests.

ddl/cache.lua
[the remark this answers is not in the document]
0x501D: But old version was better to my mind. Cache can be rebuilded by trigger, so we will get unnecessary cache rebuild.

ddl/cache.lua
```diff
@@ -0,0 +1,136 @@
...
+--                      -- or a function ready to call (or a string with an error)
+--     }
+--   },
+--   lock -- locking based on fiber.channel()
```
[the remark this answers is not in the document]
0x501D: fixed

ddl/cache.lua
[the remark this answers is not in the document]
0x501D: done

ddl/cache.lua
```diff
@@ -0,0 +1,136 @@
...
+--       processed = {}, -- table with parsed dot notation (like {'foo', 'bar'})
+--                       -- or a function ready to call (or a string with an error)
+--     }
+--   },
+--   lock,               -- locking based on fiber.channel()
+--   schema_version,     -- current schema version
+-- }
+--
+-- function returns nothing
+local cache_build = locked(function()
```
Totktonada, line 45: Initially we discussed locking, because I was not sure, whether setting of a trigger yields / considered as DDL. I had no full information here, sorry. Anyway, locking around current implementation of the `cache_build()` function (without setting a trigger) is useless: `:pairs()` does not yield for memtx, all other operations are not about the database: just Lua object changes. Unlikely `_ddl_sharding_func` will be a vinyl space. But we can wrap it into `box.atomic()` to work correctly with vinyl too (it is harmless).

What about locking around listing / setting triggers? I looked over the code and found that this action just touch some runtime information. It does not yield. So, it seems, that locking here is not necessary.

Sorry for directing you by the wrong way.

ddl/cache.lua
```diff
@@ -0,0 +1,136 @@
...
+        end
+    end
+end
+
+-- Build cache.
+--
+-- Cache structure format:
+-- cache = {
+--   spaces = {
+--     'space_name' = {
```
Totktonada, line 34: Nit: In Lua syntax it would be either `{space_name = <...>}` or `{['space_name'] = <...>}`. Or, if you want `space_name` be like a variable, `{[space_name] = <...>}`. The latter variant is used in the README.md.

ddl/cache.lua
```diff
@@ -0,0 +1,136 @@
...
+                cache.spaces[space_name].processed =
+                    string.format("Body is incorrect in sharding_func for space (%s): %s",
+                    space_name, err)
+            else
+                cache.spaces[space_name].processed =
+                    sharding_func()
+            end
+        elseif func_name ~= nil then
+            local chunks = string.split(func_name, '.')
+            cache.spaces[space_name].processed = chunks
```
Totktonada, line 74: Nit: As for me, it requires a clarification: why we don't save the function itself into the cache? I know, it is because the function can be changed in runtime and we have no way to catch this change, but anyway: another reader may be confused.

ddl/cache.lua
Totktonada: BTW, I would suggest the following style:

```lua
{
    -- A description here.
    field = <...>,
}
```

Instead of inline comments. We usually use this style and, as for me, it generally looks more accurate.

ddl/cache.lua
```diff
@@ -0,0 +1,136 @@
...
+
+-- Build cache.
+--
+-- Cache structure format:
+-- cache = {
+--   spaces = {
+--     'space_name' = {
+--       raw = {},       -- raw sharding metadata, used for ddl.get()
+--       processed = {}, -- table with parsed dot notation (like {'foo', 'bar'})
+--                       -- or a function ready to call (or a string with an error)
```
Totktonada, line 37: A kind of nitpicking or thoughts on the code. How about the following structure?

```lua
{
    error = <string> or <nil>,
    raw_tuple = <tuple object> (<nil> at error),
    callable = <function> or <nil>,
    parsed_func_name = <table> or <nil>,
}
```

It avoid polymorph fields and may be simpler to understand and use.

I don't mind current structure: it solves given task. If you disagree, feel free to close this comment.

ddl/get.lua
```diff
@@ -66,11 +67,24 @@ local function get_metadata(space_name, metadata_name)
...
 
+    return record.processed
+end
+
+local function get_sharding_func_raw(space_name)
+    local record = cache.internal.get(space_name)
+
+    if not record or not record.raw then
+        return nil
+    end
```
Totktonada, line 84: While we'll always get either tuple or `nil`, it seems natural to look, whether the cache contains an error, or explicitly clarify that the error corresponds only to `processed` and not `raw` in the cache structure description, or comment why exactly we ignore the error from the cache here.

test/cache_test.lua
```diff
@@ -0,0 +1,151 @@
...
+        body = func_body_first
+    }
+    local ok, err = ddl.set_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(ok, true)
+
+    local res = cache.internal.get('space')
+    t.assert(res)
+    t.assert(res.processed)
+    t.assert(type(res.processed) == 'function')
```
Totktonada, line 73: Nit: I suggest to use the `t.assert_type(value, expected_type)` shortcut, because in case of an error it'll show the actual type.

ddl/cache.lua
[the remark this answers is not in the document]
Totktonada: NB: AFAIU, we don't need to update the cached schema version when reacting to a change in the `_ddl_sharding_func` space: it is not DDL, so it'll not bump the schema version. It is harmless to re-read it, though. Both variants of the code looks correct and clear enough for me.

test/cache_test.lua
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env tarantool
```
Totktonada, line 1: I propose to test several more scenarious:

* A function in `_G` is changed without updating its name in the schema — verify that we don't cache the old callable.
* A function name in the schema is changed (does not matter, whether directly in the `_ddl_sharding_func` space or using `ddl.set_schema()`) — for the sake of completeness, we do that only for a function body for now.
* A space itself is changed (say, we add one more field into a sharding key definition) — it is why we look at the schema version.

Totktonada:
Since we aim to resolve a performance problem here I would like to see some performance measurements. For example, the following:

* Baseline: just call our own bucket_id function directly (I suggest to extract mpcrc32 from vshard).
* Bad case: call `ddl.bucket_id()` before this patch (name / body).
* Good case: call it after the patch (name / body).

It would be good to commit the microbenchmark right into the repository with a short instruction.

If the good case is not so good, we can look a bit deeper and, say, check, whether implementing a fast path for a function that is placed directly in `_G` will change the situation (here we don't need a loop to extract a nested field with the function). Or, maybe, we'll spot other problematic places. Or, in turn, we can see that the caching does not give anything (I would be surprised, but anyway).

test/cache_test.lua
0x501D: Done

test/cache_test.lua
0x501D: fixed

ddl/get.lua
0x501D: fixed, comment added

ddl/cache.lua
0x501D: reworked, done

ddl/cache.lua
0x501D: comment added

ddl/cache.lua
0x501D: fixed

ddl/cache.lua
0x501D: fixed, locked() removed. box.atomic is used instead

test/cache_test.lua
```diff
@@ -0,0 +1,163 @@
...
+    t.assert_equals(res.raw_tuple[SHARD_FUNC_NAME_IDX], nil)
+    t.assert_equals(res.raw_tuple[SHARD_FUNC_BODY_IDX], func_body_first)
+
+    space_init(g)
+
+    local res = cache.internal.get('space')
+    t.assert_equals(res, nil)
+end
+
+function g.test_cache_space_updated()
```
Totktonada, line 139: Shouldn't we check what the callable will return as well?

test/cache_test.lua
Totktonada: > A function in `_G` is changed without updating its name in the schema — verify that we don't cache the old callable.

`test_cache_func_body_changed()` does it. Okay.

> A function name in the schema is changed (does not matter, whether directly in the `_ddl_sharding_func` space or using `ddl.set_schema()`) — for the sake of completeness, we do that only for a function body for now.

`test_cache_space_updated()` does it for a stored function body. But what about updating a stored function name?

> A space itself is changed (say, we add one more field into a sharding key definition) — it is why we look at the schema version.

`test_cache_schema_changed()` looks similar on the first glance, but AFAIU it actually changes `_ddl_sharding_func` and effectively equivalent to `test_cache_space_updated()`.

I just observed that my example 'add one more field into a sharding key definition' is wrong: it is `_ddl_sharding_key` change (not a 'true' DDL) and, moreover, doesn't affect `ddl.bucket_id()` result, because it accepts a key, not a tuple.

I think more on it... and why we need to track schema version at all? The only source information for the cache is the `_ddl_sharding_func` content. We don't check even whether a space exists and maybe shouldn't (not within `ddl.bucket_id() at least).

ddl/cache.lua
```diff
@@ -0,0 +1,120 @@
...
+        cache_set_trigger()
+    end
+
+    -- rebuild cache if database schema changed
+    if schema_version ~= cache.schema_version then
+        box.atomic(cache_build)
+        cache_set_trigger()
+    end
+
+    return cache.spaces[space_name]
```
Totktonada, line 113: Cited from the issue:

> Don't use the cache in a transaction? That's slow, but how else we can ensure that the cache will correspond to the state visible in the transaction?

It seems, current implementation does not reflect this point. So, if we're in a transaction, which can see a more fresh state than should be visible in the transaction. It could be achieved by using vinyl for `_ddl_sharding_func` (which has no sense) or by using MVCC mode for memtx (which is possible).

However I propose to clarify that the `ddl.bucket_id()` function is not transactional in the documentation instead of fixing it with performance loss. Because the performance loss may be a real user visible problem, but I can't imagine a practical case of a sharding function, which would depend on data in some space, which will be changed together with `_ddl_sharding_func` in a transaction as part of some migration. Really strage. Let's just leave a note in docs.

However I have one more practical question here. `box.atomic()` will raise an error inside a transaction, because tarantool does not support nested transactions:

```lua
tarantool> box.atomic(function() box.atomic(function() end) end)
---
- error: 'Operation is not permitted when there is an active transaction '
...
```

So, it the current implementation, `ddl.bucket_id()` cannot be used in a transaction at all. We shouldn't wrap the cache building into `box.atomic()` if we're in a transaction. There is [box.is_in_txn()](https://github.com/tarantool/tarantool/issues/3518) since 1.10.3 and 2.1.0, so I think that we can use it unconditionally.

Dropping `box.atomic()` here at all is okay too. I doubt that anyone ever will create `_ddl_sharding_func` in vinyl. A comment in the code (that clarifies our assumptiouns) is more than enough.

I vote for dropping `box.atomic()`. If you're agree, I expect the following changes:

* Drop `box.atomic()`.
* Add a comment into the code regarding our assumptions (`_ddl_sharding_func` is memtx space, so `:pairs()` on it is atomic — AFAIU even in MVCC mode).
* Add a comment into README: `ddl.bucket_id()` is not transactional in the sense that it catches up `_ddl_sharding_func` changes immediatelly: it may see changes that're not committed yet and may see a state from another transaction, which should not be visible in the current transaction.

----

I have a raw idea how to overcome the 'read uncommited' problem (we can use on_replace() + on_commit()), but we anyway can't have a separate cache version for a transaction, which changes a space we're interested in. I propose to don't do anything with it. It is interesting idea to implement in tarantool, though: I'll file an issue regarding this later.

test/bench_cache.lua
```diff
@@ -0,0 +1,180 @@
...
+            os.exit()
+        end
+    end
+
+    print("Done")
+    os.exit()
+end
+
+local function main()
+    local getopt = require('posix.unistd').getopt
```
Totktonada, line 157: Is it from luaposix? It is not something that we have installed by default. I would prefer to don't depend on it.

We can either look over `arg` ourself (here it is quite simple) or use `internal.argparse` (which is internal, sure, but it changes rarely and built-in).

test/bench_cache.lua
```diff
@@ -0,0 +1,180 @@
...
+            run_baseline()
+        elseif opt == 'n' then
+            run_name()
+        elseif opt == 'b' then
+            run_body()
+        end
+    end
+end
+
+main()
```
Totktonada, line 180: My personal experience suggests that when a test (including a benchmark) is not run in CI somehow, it'll become broken after some time. Feel free to ignore. You have been warned.

test/bench_cache.lua
0x501D: Ok, I will rewrite with manually args parsing. But I prefer using getopt() where possible.

test/cache_test.lua
0x501D: agree

test/bench_cache.lua
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env tarantool
+
+local db = require('test.db')
+local ddl = require('ddl')
+local helper = require('test.helper')
+
+local ITERS = 10000000
+
+local function mpcrc32(shard_key)
```
Totktonada, line 9: It looks more like `strcrc32()`.

test/cache_test.lua
Totktonada: After voice discussion with Pavel: we should catch the situation like the following:

```lua
box.space._ddl_sharding_func:drop()
box.schema.create_space('_ddl_sharding_func', <...>)
box.space._ddl_sharding_func:create_index(<...>)
```

We can either look at the schema version (as it is implemented now) or set a trigger on `_space`.

Look at schema version: extra rebuilds at other changes, using of internal API.

Trigger on `_space`: it is not clear for me, whether we always have proper permissions to set the trigger.

So it is okay for me to keep the schema version approach. But I vote for a comment in the code that clarifies why exactly we need to look at the schema version and for rewritting the test in more clear way: so it'll be obvious that we're about the case described above.

test/cache_test.lua
0x501D: This test case just check that body was changed after space updated (trigger works).

test/bench_cache.lua
0x501D: fixed

test/bench_cache.lua
0x501D: moved to CI

test/bench_cache.lua
0x501D: fixed

ddl/cache.lua
0x501D: box.atomic() dropped, сomment added, bucket_id() descripton in REAMDE added

.github/workflows/test.yml
```diff
@@ -66,3 +66,31 @@ jobs:
...
       - run: tarantoolctl rocks remove ddl
+  benchmark:
+    strategy:
+      fail-fast: false
+      matrix:
+        tarantool: ['1.10', '2.5', '2.6', '2.7']
+        coveralls: [false]
+        include:
+          - tarantool: '2.8'
+            coveralls: true
```
Totktonada, line 77: This is dead code (and I think that we don't need to collect coverage on the benchmark).

.github/workflows/test.yml
```diff
@@ -66,3 +66,31 @@ jobs:
...
+        with:
+          tarantool-version: ${{ matrix.tarantool }}
+
+      # Setup luatest
+      - name: Cache rocks
+        uses: actions/cache@v2
+        id: cache-rocks
+        with:
+          path: .rocks/
+          key: cache-rocks-${{ matrix.runs-on }}-06
```
Totktonada, line 91: The key is the same as in the job above, but stored `.rocks` content is different. I imagine the following situation: we store `.rocks` with luatest here and then next 'test' job will find the cache and use it. So it'll skip installation of luacheck, luacov and other modules and will fail.

Let's use another key.

Totktonada: BTW, as I see from the benchmark code, it does not depend on luatest. Is it so?

.github/workflows/test.yml
```diff
@@ -66,3 +66,31 @@ jobs:
...
+      -
+        run: tarantoolctl rocks install luatest
+        if: steps.cache-rocks.outputs.cache-hit != 'true'
```
Totktonada, line 94: Hm. Unusual formatting is here. I didn't know that it works this way. I would reformat it to:

```yaml
- run: <...>
  if: <...>
```

ddl/cache.lua
Totktonada: Okay.

Filed https://github.com/tarantool/tarantool/discussions/6800 with a dump of my current thoughts.

.github/workflows/test.yml
0x501D: depends from local helper = require('test.helper') -> local helpers = table.copy(require('luatest').helpers)

test/cache_test.lua
0x501D: test added and fixed

test/cache_test.lua
0x501D: fixed

.github/workflows/test.yml
0x501D: deleted

.github/workflows/test.yml
0x501D: fixed

.github/workflows/test.yml
0x501D: fixed

README.md
```diff
@@ -89,6 +89,11 @@ bucket identifier (number)
     - Calculate bucket id for a specified space and sharding key.
     Method uses sharding function specified in DDL schema.
 
+    Method is not transactional in the sense that it catches up
+    _ddl_sharding_func changes immediatelly: it may see changes that're
```
ligurio, line 93: Nit: Usually, we mark words related to source code with apostrophes for better visibility.
This one `_ddl_sharding_func` is looking better than "_ddl_sharding_func". However, it may look good for your taste, so I will not insist.

test/bench_cache.lua
```diff
@@ -0,0 +1,168 @@
...
+    if type(shard_key) ~= 'table' then
+        return digest.crc32(tostring(shard_key))
+    else
+        local crc32 = digest.crc32.new()
+        for _, v in ipairs(shard_key) do
+            crc32:update(tostring(v))
+        end
+        return crc32:result()
+    end
+end
```
ligurio, line 21: The same code is existing in `test/helper.lua`. I would reuse it instead of copy-pasting.

test/bench_cache.lua
```diff
@@ -0,0 +1,168 @@
...
+]]
```
ligurio, line 36: It would be better to reuse variable from `test/helper.lua`

ddl/cache.lua
```diff
@@ -0,0 +1,124 @@
...
+--       parsed_func_name = <table> or <nil>
+--       -- a function ready to call
+--       callable = <function> or <nil>,
+--       -- string with an error: not nil only if setting callable fails
+--       error = <string> or <nil>,
+--     }
+--   },
+--   -- current schema version
+--   schema_version = <...>,
+-- }
```
ligurio, line 30: Nit: Two spaces instead of four used in indentation.

test/cache_test.lua
```diff
@@ -0,0 +1,245 @@
...
+    rawset(_G, sharding_func_first, function(key) return key + 1 end)
+    rawset(_G, 'sharding_func_second', function(key) return key + 2 end)
+    g.schema.spaces.space.sharding_func = sharding_func_first
+    local ok, err = ddl.set_schema(g.schema)
+    t.assert_equals(err, nil)
+    t.assert_equals(ok, true)
+
+    local res = cache.internal.get()
+    t.assert_equals(res, nil)
+end
```
ligurio, line 245: Added a small testcase to cover a case when space name is `nil`. Test makes a 100% coverage for module `ddl/cache.lua`. 

Screenshot from 2022-01-25 12-53-01


```lua
--- a/test/cache_test.lua
+++ b/test/cache_test.lua
@@ -230,3 +230,16 @@ function g.test_cache_space_updated_name()
     t.assert_type(id, 'number')
     t.assert_equals(id, 44)
 end
+
+function g.test_cache_nil_space()
+    local sharding_func_first = 'sharding_func_first'
+    rawset(_G, sharding_func_first, function(key) return key + 1 end)
+    rawset(_G, 'sharding_func_second', function(key) return key + 2 end)
+    g.schema.spaces.space.sharding_func = sharding_func_first
+    local ok, err = ddl.set_schema(g.schema)
+    t.assert_equals(ok, true)
+
+    local res = cache.internal.get()
+    t.assert_equals(res, nil)
+end
```

ligurio (approved):
Thanks for your fixes!

In general everything is fine, however I left several non-critical comments. Please consider them. I will give an LGTM to don't block merge.

test/cache_test.lua
Totktonada: (Or we can remove the check — it is anyway not a valid input and I hope that API functions doing input validation good enough.)

ddl/cache.lua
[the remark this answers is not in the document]
Totktonada: (It seems, it is not actual anymore. Resolving the discussion.)

ddl/cache.lua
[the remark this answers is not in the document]
Totktonada: (It seems, it is not actual anymore. Resolving the discussion.)

ddl/cache.lua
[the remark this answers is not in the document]
Totktonada: (It seems, it is not actual anymore. Resolving the discussion.)

ddl/cache.lua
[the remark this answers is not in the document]
Totktonada: (It seems, it is not actual anymore. Resolving the discussion.)

ddl/get.lua
Totktonada: (It seems, it is not actual anymore. Resolving the discussion.)

test/cache_test.lua
Totktonada: (It seems, it is not actual anymore. Resolving the discussion.)

test/cache_test.lua
Totktonada: (It seems, it is not actual anymore. Resolving the discussion.)

test/cache_test.lua
Totktonada: (It seems, it is not actual anymore. Resolving the discussion.)

test/cache_test.lua
Totktonada: I'm satisfied with the test cases now.

Totktonada: Thanks!

Totktonada (approved):
LGTM.

Please, address stylistic suggestions from Sergey, squash (or rework-and-squash) the proposed commit into your one and let me know that everything is done.

test/cache_test.lua
0x501D: I prefer insurance, test added

ddl/cache.lua
0x501D: fixed

test/bench_cache.lua
0x501D: fixed

test/bench_cache.lua
0x501D: fixed

README.md
0x501D: fixed

Totktonada: (Looks resolved, so I'm going to resolve this discussion.)

ddl/cache.lua
Totktonada: (Looks resolved, so I'm going to resolve this discussion.)

test/cache_test.lua
Totktonada: (Looks resolved, so I'm going to resolve this discussion.)

test/bench_cache.lua
Totktonada: (Looks resolved, so I'm going to resolve this discussion.)

test/bench_cache.lua
Totktonada: (Looks resolved, so I'm going to resolve this discussion.)

PR #95, opened 2022-07-11
test: shuffle tests

ligurio:
luatest has an option --shuffle that allows to randomise the order of
tests. This can be useful to detect a test that passes just because it
happens to run after an unrelated test that leaves the system in a
favourable state. Patch enable tests shuffling with predefined random
seed generated by CMake [1]. Seed is useful for reproducing a problems
related to the specific test order.

https://cmake.org/cmake/help/latest/command/string.html#random

### DIFF
--- a/test/set_sharding_metadata_test.lua
+++ b/test/set_sharding_metadata_test.lua
@@ -154,8 +154,12 @@ function g.test_two_sharding_spaces()
     space_two.sharding_key = {
         'unsigned_nonnull', 'integer_nonnull', 'string_nonnull'
     }
-    local sharding_func_name = 'some_module.sharding_func'
-    rawset(_G, sharding_func_name, function(key) return key end)
+    local some_module = {
+        sharding_func = function(key) return key end
+    }
+    local sharding_func_module = 'some_module'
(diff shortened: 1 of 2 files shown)


test/set_sharding_metadata_test.lua
```diff
@@ -154,8 +154,11 @@ function g.test_two_sharding_spaces()
     space_two.sharding_key = {
         'unsigned_nonnull', 'integer_nonnull', 'string_nonnull'
     }
+    local some_module = {
```
olegrok, line 157: To be honest I don't understand how it could fix anything? Please describe.

olegrok (approved):

test/set_sharding_metadata_test.lua
olegrok: Ok. Thanks for explanation.

But if there is the same code in different test you can simply move this logic to before_tests hook

olegrok: Ok. Anyway lgtm.

AnaNek (approved):
LGTM!

PR #102, opened 2022-11-17
Fix usage of Tarantool internal API for retrieving schema version

### LINKED ISSUE #101: Start using `box.info.schema_version` instead of `box.internal.schema_version`
tarantool/tarantool#7904 exports schema version to public API and Lua:
start using it, when available, instead of unreliable internal API.

CuriousGeorgiy:
tarantool/tarantool#7904 exports schema version to public API and Lua: start using it, when available, instead of unreliable internal API.

Closes #101

### DIFF
--- a/ddl/cache.lua
+++ b/ddl/cache.lua
@@ -65,7 +65,8 @@ local function cache_build()
         end
     end
 
-    cache.schema_version = box.internal.schema_version()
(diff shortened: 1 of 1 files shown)


AnaNek (approved):
LGTM! Thank you!

PR #105, opened 2023-03-07
api: add hardcoded versioning support

GRISHNOV:
Add versioning support, now the module api has `_VERSION` hardcoded variable. Is part of the task [1].

1. https://github.com/tarantool/roadmap-internal/issues/204


GRISHNOV:
note: `Test/test` in ci/cd have a problem, it is not caused by the current task

DifferentialOrange (approved):

AnaNek (approved):

PR #106, opened 2023-03-25
ci: workaround setup-tarantool problem with Jammy

GRISHNOV:
There is known unsolved problem with Ubuntu Jammy [1].  Because of this, `tarantool 2.6` is installed in all cases. During tests it turns out that `builtin#95 table.clear` is [nil](https://github.com/tarantool/ddl/actions/runs/4518349665/jobs/7958166174#step:15:3) on the installed tarantool 2.6, which causes problems in ci job `test`.

Also the expected tarantool version was not installed in job `benchmark` before this patch.

1. tarantool/setup-tarantool/issues/36

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,13 @@ jobs:
         include:
           - tarantool: '2.8'
             coveralls: true
-    runs-on: [ubuntu-latest]
+    # There are problems with current version of the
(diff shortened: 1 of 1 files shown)


AnaNek (approved):
LGTM. Thank you very much for this patch!

DifferentialOrange (approved):

PR #114, opened 2023-08-17
test: align `set_schema_test.lua` with core `exclude_null` changes

Gumix:
The fix for tarantool/tarantool#8649 always sets `index_object.parts.exclude_null` to `false` when it is actually `false`, however the test `set_schema_test.lua` fails [1], because it expects `index_object.parts` without the `exclude_null` field. This patch fixes the test so that it allows both `nil` and `false` for `exclude_null`.

1. https://github.com/tarantool/tarantool/actions/runs/5889232292/job/15972202950

### DIFF
--- a/test/set_schema_test.lua
+++ b/test/set_schema_test.lua
@@ -1327,9 +1327,15 @@ g.test_gh_108_fieldno_index_no_space_format = function()
     t.assert_equals(err, nil)
 
     t.assert_equals(box.space['weird_space']:format(), {})
-    t.assert_equals(
-        box.space['weird_space'].index['pk'].parts,
-        {{fieldno = 1, type = 'unsigned', is_nullable = false}})
+
+    -- Tarantool may return is_nullable as `nil` or as `false`, depending on the
+    -- version (see tarantool/tarantool#8649 for details). Allow both variants.
+    local expected_variants = {
+        {{fieldno = 1, type = 'unsigned', is_nullable = false}},
+        {{fieldno = 1, type = 'unsigned', is_nullable = false,
+          exclude_null = false}}}
+    t.assert_items_include(expected_variants,
+                           {box.space['weird_space'].index['pk'].parts})
 
     t.assert_equals(box.space['weird_space']:insert{1}, {1})
     t.assert_equals(box.space['weird_space']:insert{2, 'val'}, {2, 'val'})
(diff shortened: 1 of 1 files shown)


test/set_schema_test.lua
```diff
@@ -1354,9 +1359,14 @@ g.test_gh_108_fieldno_index_in_space_format = function()
 
     t.assert_equals(box.space['weird_space']:format(),
         {{is_nullable = false, name = "id", type = "unsigned"}})
-    t.assert_equals(
-        box.space['weird_space'].index['pk'].parts,
-        {{fieldno = 1, type = 'unsigned', is_nullable = false}})
+
+    local res = box.space['weird_space'].index['pk'].parts
+    if res[1].exclude_null == nil then
```
DifferentialOrange, line 1364: Let's add a comment here (and below, copypasting here shouldn't be bad for anyone) with the same contents as your commit message so it would be obvious what's going on.

test/set_schema_test.lua
```diff
@@ -1327,9 +1327,14 @@ g.test_gh_108_fieldno_index_no_space_format = function()
...
     t.assert_equals(box.space['weird_space']:format(), {})
-    t.assert_equals(
-        box.space['weird_space'].index['pk'].parts,
-        {{fieldno = 1, type = 'unsigned', is_nullable = false}})
+
+    local res = box.space['weird_space'].index['pk'].parts
+    if res[1].exclude_null == nil then
+        res[1].exclude_null = false
+    end
+    t.assert_equals(res,
```
DifferentialOrange, line 1335: I prefer `assert_items_include` for branching results. It seems like a hack, but still better than overwriting test response.

```lua
tarantool> t = require('luatest')
---
...

tarantool> variant_1 = {{fieldno = 1, type = 'unsigned', is_nullable = false,
          exclude_null = false}}
---
...

tarantool> variant_2 = {{fieldno = 1, type = 'unsigned', is_nullable = false,
          exclude_null = nil}}
---
...

tarantool> variant_3 = {{fieldno = 1, type = 'unsigned', is_nullable = false,
          exclude_null = true}}
---
...

tarantool> allowed_variants = {variant_1, variant_2}
---
...


tarantool> t.assert_items_include(allowed_variants, {variant_1})
---
...

tarantool> t.assert_items_include(allowed_variants, {variant_2})
---
...

tarantool> t.assert_items_include(allowed_variants, {variant_3})
---
- error:
    status: fail
    class: LuatestError
    message: "Expected all item values from: \n{{{exclude_null = true, fieldno = 1,
      is_nullable = false, type = \"unsigned\"}}}\nTo be present in: \n{\n    {{exclude_null
      = false, fieldno = 1, is_nullable = false, type = \"unsigned\"}},\n    {{fieldno
      = 1, is_nullable = false, type = \"unsigned\"}},\n}"
...
```

DifferentialOrange (changes requested):
Thank you for your PR! The idea is understandable, but I think it's better to rework a couple of things.

test/set_schema_test.lua
Gumix: Done.

test/set_schema_test.lua
Gumix: Thanks for the hint, done.

PR #115, opened 2023-09-13
test: align set_schema_test.lua with sort_order feature

mkostoevr:
The pull request tarantool/tarantool#8915 implements the key part sort order, which introduces a new `sort_order` field of index parts. This field is unexpected for the test and so it fails [1].

This patch fixes the test by allowing to have this part property.

1. https://github.com/tarantool/tarantool/actions/runs/6172230004/job/16752337479


DifferentialOrange:
Is it ready for review?

mkostoevr:
@DifferentialOrange, shure.

PR #117, opened 2023-10-23
ddl: support interval type in format

DifferentialOrange:
Datetime intervals are supported as a part of space schema format since 2.10.0 [1]. They are not allowed to be a part of memtx index [2].

1. https://github.com/tarantool/tarantool/commit/a9e30fb365c1643484054aa07038c015383e7faa
2. https://github.com/tarantool/tarantool/blob/ddaa5a320ee8d8d1009e07bbf6b8fe5cb4acbb45/src/box/memtx_space.c#L844-L851

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -36,11 +36,19 @@ jobs:
         with:
           tarantool-version: '2.11'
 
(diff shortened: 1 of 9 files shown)


better0fdead (approved):
Hi, lgtm

PR #126, opened 2024-05-28
test: fix error message testes after Tarantool

nshy:
We stop to add 'Illegal parameters, ' in https://github.com/tarantool/tarantool/pull/10058.


DifferentialOrange (approved):
testes -> tests

PR #129, opened 2025-04-03
ci: bump ubuntu version

patapenka-alexey:
Bump actions to use ubuntu-24.04 for fixing the following GitHub warning:

    The Ubuntu 20.04 Actions runner image will begin deprecation
    on 2025-02-01.

Update list of `tarantool` versions to test.

Part of #TNTP-1918

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -9,7 +9,7 @@ jobs:
   version-check:
     # We need this job to run only on push with tag.
     if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Check module version
         uses: tarantool/actions/check-module-version@master
@@ -18,7 +18,7 @@ jobs:
 
   publish-scm-1:
     if: github.ref == 'refs/heads/master'
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
       - uses: tarantool/rocks.tarantool.org/github-action@master
@@ -29,15 +29,15 @@ jobs:
   publish-tag:
     if: startsWith(github.ref, 'refs/tags/')
     needs: version-check
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
-      - uses: tarantool/setup-tarantool@v2
+      - uses: tarantool/setup-tarantool@v3
         with:
           tarantool-version: '2.11'
 
       - name: Prepare apt repo
-        run: curl -L https://tarantool.io/release/2/installer.sh | bash
+        run: curl -L https://tarantool.io/release/3/installer.sh | bash
 
       - name: Install tt cli
         run: sudo apt install -y tt
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,27 +9,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        tarantool: ['1.10', '2.5', '2.6', '2.7', '2.8', '2.10']
-        coveralls: [false]
-        include:
-          - tarantool: '2.11'
-            coveralls: true
-    # There are problems with current version of the
-    # setup-tarantool action on Ubuntu Jammy (ubuntu-latest or
-    # ubuntu-22.04). Use Ubuntu Focal (ubuntu-20.04) until they
-    # will be resolved. See [1] for details.
-    #
-    # [1]: https://github.com/tarantool/setup-tarantool/issues/36
-    runs-on: [ubuntu-20.04]
+        tarantool: ['2.11', '3.3']
+
+    runs-on: [ubuntu-24.04]
     steps:
       - uses: actions/checkout@v4
 
-      - uses: tarantool/setup-tarantool@v2
+      - uses: tarantool/setup-tarantool@v3
         with:
           tarantool-version: ${{ matrix.tarantool }}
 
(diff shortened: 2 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -66,6 +61,7 @@ jobs:
 
       - name: Run tests and code coverage analysis
         run: make -C build coverage
+        if: ${{ matrix.tarantool != '3.3' }}
```
patapenka-alexey, line 64: Could we use such type of workarounds if coverage fails on `3.3.1`?

.github/workflows/publish.yaml
```diff
@@ -29,15 +29,15 @@ jobs:
...
     needs: version-check
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
-      - uses: tarantool/setup-tarantool@v2
+      - uses: tarantool/setup-tarantool@v3
         with:
-          tarantool-version: '2.11'
+          tarantool-version: '1.10'
```
elhimov, line 37: This change looks like an alien among others upgrades/updates/bumps/etc. Is this downshifting really necessary?

elhimov (changes requested):

.github/workflows/publish.yaml
patapenka-alexey: hm, let's check on CI.

.github/workflows/test.yml
elhimov: I'm not familiar with ddl but generally speaking it doesn't look like good idea. What kind of failure occurs with 3.3.1?

elhimov (approved):
LGTM, except for the coverity issue. Need more confident and qualified answer.

oleg-jukovec (approved):

.github/workflows/test.yml
oleg-jukovec: Why don't we count coverage for 3.3?

oleg-jukovec, line 64:
```suggestion
        if: ${{ matrix.tarantool == '3.3' }}
```

.github/workflows/test.yml
```diff
@@ -9,22 +9,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        tarantool: ['1.10', '2.5', '2.6', '2.7', '2.8', '2.10']
+        tarantool: ['2.11', '3.3']
```
elhimov, line 12: Originally matrix resulted 7 jobs (coveralls=true for 2.11 and false for the rest ones). Now it results in - 2.11 + false
- 3.3 + false
- 2.11 + true
Do we really need this set? Do we understand the meaning of coveralls? Do we know the reason of the original exception for 2.11? I don't know the meaning of coveralls, but it doesn't look necessary to test both coveralls (true+false) for 2.11.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.05...3.31 FATAL_ERROR)
```
oleg-jukovec, line 1:
```suggestion
cmake_minimum_required(VERSION 3.10...3.31 FATAL_ERROR)
```

.github/workflows/test.yml
patapenka-alexey: Changed to `matrix.coveralls`

CMakeLists.txt
patapenka-alexey: Fixed.

test/tarantool_helpers.lua
oleg-jukovec, line 1: Please, remove unused functions.

oleg-jukovec (changes requested):

test/tarantool_helpers.lua
patapenka-alexey: Removed.

.github/workflows/test.yml
```diff
@@ -66,6 +61,7 @@ jobs:
 
       - name: Run tests and code coverage analysis
         run: make -C build coverage
+        if: ${{ matrix.coveralls }}
```
elhimov, line 64: Please, correct if I'm wrong but according to names of the steps it seems coveralls means just whether to send or not the result of coverage analysis. New approach means don't perform coverage analysis at all if not going to send it. Is it the desired workflow? Just to make sure ))

patapenka-alexey: > Is it the desired workflow?

We should disable coverage tests build and running for Tarantool 3.x, so I think this is the easiest way to do it.

elhimov: But this way it looks like a kind of cryptography, because our intention is well hidden )
If we should disable coverage tests for 3.x, why don't we just put here something like
```
if: matrix.tarantool == "3.3"
```
or
```
if: startsWith(matrix.tarantool, "3.")
```
?

.github/workflows/test.yml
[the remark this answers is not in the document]
patapenka-alexey: It works on my machine, but ok - updated.

test/role_test.lua
```diff
@@ -4,8 +4,11 @@ local g = t.group()
 local fio = require('fio')
 local yaml = require('yaml')
 local helpers = require('test.helper')
+local tarantool_helpers = require('test.tarantool_helpers')
 
 g.before_all(function()
+    tarantool_helpers.skip_if_tarantool3()
```
oleg-jukovec, line 10: It should be enough. You could delete other skips in the file.

oleg-jukovec (changes requested):

elhimov (approved):

oleg-jukovec (approved):

